PR #78, opened 2023-12-02
Add save button + saved properties display in side panel

nlebovits:
Not a perfect PR--still some small styling issues with the buttons and stuff. But the important stuff is there. Lmk how we can resolve merge conflicts and I'll take care of it.


brandonfcohen1:
to be perfectly honest- since this is a pretty small feature it might just be worth you re-writing from the updated `main`. These conflicts are pretty deep

brandonfcohen1 (changes requested):
recommend re-writing from current `main`

Issue #94, opened 2023-12-05
menu bar icons get small on narrow screens


brandonfcohen1:
![image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/2308834/13e4b510-6a80-4abf-9440-a513ff94e6ef)

keep icon size consistent regardless of screen width

jroper-scottlogic:
HI, I made a pr to hopefully fix this here
https://github.com/CodeForPhilly/vacant-lots-proj/pull/95

Issue #108, opened 2023-12-07
Task: remove searchbar from MapControlBar.tsx


nlebovits:
Now that the geocoder search bar has been integrated into the PropertyMap.tsx via react-map-gl, we can remove the search bar from the MapControlBar.tsx component.

gramzanali-scottlogic:
Hi, I've opened a pr that should close this issue here [#115](https://github.com/CodeForPhilly/vacant-lots-proj/pull/115).

PR #115, opened 2023-12-12
Remove Searchbar From MapControlBar.tsx

gramzanali-scottlogic:
Closes [Issue 108](https://github.com/CodeForPhilly/vacant-lots-proj/issues/108) by removing the code associated with the search bar from the MapControlBar.tsx component.

Before: 
/map page with the 2 search bars visible

After:
/map page with 1 search bar visible

### DIFF
--- a/src/app/components/MapControlBar.tsx
+++ b/src/app/components/MapControlBar.tsx
@@ -1,11 +1,7 @@
-import React, { useState } from "react";
-import { Input, Button } from "@nextui-org/react";
+import React from "react";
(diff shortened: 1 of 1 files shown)


nlebovits:
Nice, @gramzanali-scottlogic! Thanks for the contribution! @brandonfcohen1 if it looks good to you, we can merge it. Frees up some space for the save button and maybe will also fix #94?

brandonfcohen1 (approved):
looks good thanks!

Issue #103, opened 2023-12-06
Create save button to store selected properties locally


nlebovits:
People searching the map may want to save a selection of properties to look at later. Storing with state will result in these properties being lost when a user refreshes the page. Set up a save button to store this list of properties locally so that the user can return to them later.

nlebovits:
> we can use `localStorage`. Might have to add one of those GDPR notices if we do 🤷‍♂️

Yeah, that's what I was getting add. Good thinking about the notice. This is low priority anyway.

nlebovits:
stale--new ticket out

PR #116, opened 2023-12-18
add saved button w local storage

nlebovits:
Save button working with local storage. Some typescript issues to clean up still, but I needed to get this in for a final project.


nlebovits:
closes #103

brandonfcohen1 (changes requested):
nice work.

problem: if I click on a saved property, it disappears

we should also add a way to save from the grid view or table view

nlebovits:
@brandonfcohen1 noted. will leave until later--not high priority, so we can come back to it.

PR #122, opened 2023-12-21
Add OPA properties data to dataset

nlebovits:
Adding OPA properties data to dataset to be integrated into access process conditional logic


brandonfcohen1:
can you fix merge conflicts @nlebovits ?

brandonfcohen1 (changes requested):
remove saved properties branch work and re-submit

Issue #117, opened 2023-12-19
Task: set up "all contributors" bot to document contributions in github repo


nlebovits:
To better recognize contributors to the project, we should install and configure the [All Contributors bot](https://allcontributors.org/docs/en/bot/installation) on this GitHub repo, and also add contributors from the UX team who have contributed to things other than code.

nlebovits:
App has been added to the GH account. Only steps 3 and 4 now needed: https://allcontributors.org/docs/en/bot/installation.

nlebovits:
@all-contributors please add @nlebovits for getting this running!

nlebovits:
@all-contributors please add @nlebovits for documentation!

PR #125, opened 2023-12-21
Get all contributors bot working

nlebovits:
Should address #117


brandonfcohen1 (changes requested):
get rid of your saved button commit :)

Issue #142, opened 2024-01-09
Task: Add tactical urbanism column


nlebovits:
Add a column to the dataset to indicate properties that are suitable for tactical urbanism. These will be all properties that are:

1) vacant land, not a building AND
2) not unsafe or imminently dangerous

For part 1, the easiest way to go about this is to string match for VAC LAND in the building_desc column from the OPA dataset.

For part 2, you'll need to add both the unsafe and imminently dangerous datasets from here: https://opendataphilly.org/datasets/licenses-and-inspections-code-violations/. This can be done using a join on the opa_id column rather than a spatial join.

PR #176, opened 2024-01-15
163 fix access process script

brandonfcohen1:
@nlebovits I believe I fixed this, take a look. here are the final counts:
image

### DIFF
--- a/data/src/config/config.py
+++ b/data/src/config/config.py
@@ -1,5 +1,5 @@
 import os
 
-FORCE_RELOAD = True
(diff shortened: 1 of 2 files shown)


nlebovits (approved):
Code looks right to me. Total counts are interesting. I'll dig some more--maybe I need to fine-tune the access process issues. Can you add the tactical urbanism column while you're at it? #142. Should give us more insight into what can be done with all those "Private Land Use Agreement" properties.

Issue #88, opened 2023-12-03
Task: When clicking on a property on the map, load it in the side view


brandonfcohen1:
## Describe the task
When clicking on a property on the map, load it in the side view

markgoetz:
@brandonfcohen1 Question for you - would we want to make the side view page its own URL, so that it can be deeplinked?  This would enable individual properties to be bookmarked and returned to individually.

nlebovits:
> @brandonfcohen1 Question for you - would we want to make the side view page its own URL, so that it can be deeplinked? This would enable individual properties to be bookmarked and returned to individually.

I don't think this is a bad idea. See #103 as well. Would be great to kill two birds w one stone if possible.

PR #183, opened 2024-01-17
Issue 88: opens sidebar from map

markgoetz:
Addresses #88.

To test: go to the Map page, then click on a property.  You should see the corresponding property open in the right sidebar.

OPEN QUESTION: Should we still show the popup?


nlebovits:
@all-contributors please add @markgoetz for code :)

Issue #185, opened 2024-01-17
Add New Logo


thansidwell:
As user I want to see a nice professional logo naming the website and illustrating it's value. 

1. Add the new logo as in the prototype https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=187%3A12709&mode=design&t=h1fTrSpJsuOpTu9c-1
2. Here's the logo compressed here. [Logo - Clean & Green Philly.svg.zip](https://github.com/CodeForPhilly/vacant-lots-proj/files/13958352/Logo.-.Clean.Green.Philly.svg.zip)

3. Where the logo is added in the header make sure the width is 112px and height around 67px.
4. Make sure it's not stretched or distorted.

thansidwell:
Oh, I forgot that we should take out the tagline "Clean & green vacant properties to reduce gun violence." After thinking about it a lot I think it's redundant and not necessary.

PR #190, opened 2024-01-17
add new logo

nlebovits:
added logo with 16px padding on the top margin. does it look right?
Screenshot 2024-01-16 at 23-13-17 Clean Green Philly


thansidwell:
Oh, I forgot to take out the tagline "Clean & green vacant properties to reduce gun violence." After thinking about it a lot I think it's redundant and not necessary. Other than that it looks good. I'll write another ticket with more updates to the header nav, icons, and padding.

nlebovits:
closes #185

PR #275, opened 2024-01-24
clean up homepage headline

nlebovits:


nlebovits:
@brandonfcohen1 any idea why it's failing the check?

Issue #313, opened 2024-02-05
Fix Detroit Future City quiz link


bacitracin:
**Describe the bug**
Currently appears as text and not a link:

<img width="874" alt="Screenshot 2024-02-05 at 10 27 27" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/7329799/9165b95e-c8e9-4edd-b296-6e873c18c01f">

AZBL:
Hi I just took care of this.

AZBL:
Did it again but opened to staging this time.

nlebovits:
thanks @AZBL! closed by #314

PR #314, opened 2024-02-06
fix detroit future city quiz link

AZBL:
#313 
-Fixed Detroit Future City quiz link so now it's a functional link.
-On initial commit accidentally changed formatting of Soil Generation link so this was restored to original


nlebovits:
@all-contributors please add @AZBL for code and a11y

nlebovits:
reopn

Issue #204, opened 2024-01-18
A11y - Update pagination nav's aria-label


bacitracin:
## Describe the task
The aria-label for the pagination nav is currently "pagination navigation" which announces on screenreaders redundantly - as "pagination navigation navigation."

## Acceptance Criteria
- [ ] Change the aria-label to just "pagination"

## Additional context
<img width="463" alt="Screenshot 2024-01-17 at 21 13 49" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/7329799/fccc2b88-71ea-44db-b8ad-b42092376a4f">

marvieqa:
Note: we can omit the `role=navigation` because `<nav>` has an implicit role of navigation.

markgoetz:
@bacitracin Can this issue be closed out now?

bacitracin:
@markgoetz Yeah I think we're just behind on updating the board. I'll quickly check the ones you merged last week and close them out

PR #321, opened 2024-02-09
feat: Update navigation accessibility

markgoetz:
This PR adjusts the accessibility of the navigation per #204.

### DIFF
--- a/src/app/components/PropertyDetailSection.tsx
+++ b/src/app/components/PropertyDetailSection.tsx
@@ -113,6 +113,8 @@ const PropertyDetailSection: FC<PropertyDetailSectionProps> = ({
           <div>
             <div className="flex w-full justify-center">
               <Pagination
+                role={undefined}
+                aria-label="pagination"
                 isCompact
                 showControls
                 showShadow


src/app/components/PropertyDetailSection.tsx
```diff
@@ -113,6 +113,8 @@ const PropertyDetailSection: FC<PropertyDetailSectionProps> = ({
           <div>
             <div className="flex w-full justify-center">
               <Pagination
+                role={undefined}
```
bacitracin, line 116: I'm not too familiar with the Pagination component we're using so I'm curious, did you have to pass a role to overwrite role in the component?

markgoetz: Yes; it provides "navigation" by default.

bacitracin: Wow that's silly

bacitracin (approved):
Looks great!

src/app/components/PropertyDetailSection.tsx
markgoetz: Agreed.  I wonder if it's worth filing an issue with NextUI, because I agree that the default semantics are incorrect.

Issue #298, opened 2024-01-30
A11y: "Find Properties" should be marked up as a link


marvieqa:
## Description

"Find Properties" is currently a button. It should be marked up as a link to preserve its semantic meaning since it takes you to a different URL.

<img width="1217" alt="Screenshot 2024-01-30 at 16 11 19" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/102739972/fec2f585-f906-4c49-9372-a23a5c6b4c46">

PR #323, opened 2024-02-09
feat: Change Find Properties and Learn More to links

markgoetz:
Per #298, this PR changes Find Properties to a link instead of a button nested inside of a link.  I also addressed the Learn More button because it was right there as well.

### DIFF
--- a/src/app/components/LandingPage.tsx
+++ b/src/app/components/LandingPage.tsx
@@ -37,21 +37,19 @@ const LandingPage = () => (
           </p>
         </div>
         <div className="inline-flex space-x-2">
-          <Link href="/map">
-            <Button size="lg" className="bg-green-600 text-white">
-              <ArrowRightIcon className="w-5 h-5 mr-2" />
-              Find properties
-            </Button>
-          </Link>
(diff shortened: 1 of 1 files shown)


src/app/components/LandingPage.tsx
```diff
@@ -37,21 +37,19 @@ const LandingPage = () => (
...
-          <Link href="/about">
-            <Button
-              size="lg"
-              className="bg-transparent border border-green-600 text-green-600"
-            >
-              <ArrowRightIcon className="w-5 h-5 mr-2" />
-              Learn more
-            </Button>
-          </Link>
+          <Button href="/map" as={Link} size="lg" className="bg-green-600 text-white">
```
bacitracin, line 40: I got confused when I first saw you removed Link above. But the button turns into a link 🤯

bacitracin (approved):
Keyboard nav is much nicer now!

src/app/components/LandingPage.tsx
markgoetz: It's sneaky like that!

Issue #278, opened 2024-01-24
Bug: Sometimes pages don't load


thansidwell:
About 10% of the time, when I try to navigate to these pages, they don't load.  
- Get Access
- Transform a Property
- Summary

Here's a video of it happening. https://share.cleanshot.com/CSxknnwB

It happens in both Safari and Chrome. It's not a high priority. It's hard to replicate. I'm not sure what's causing it.

nlebovits:
I've been having the same issue. I think it ties back to when I changed the menu option to a dropdown. I'm using a nextui dropdown selector but I'm wondering if three's a better way to do it. It seems like if you click on the left side of the menu subitem, it works fine, but other times it doesn't load. Not sure what the deal is, though. Maybe @markgoetz or @brandonfcohen1 have suggestions for now to refactor? issue is in Header.tsx, I think:

```
      <Dropdown>
        <DropdownTrigger>
          <Button size="md" className="flex items-center hover:text-blue-500 hover:underline bg-white">
            <Hand className="h-6 w-6" aria-hidden="true" />
            Take Action
          </Button>
        </DropdownTrigger>
        <DropdownMenu aria-label="Take Action">
          <DropdownItem key="get-access">
            <Link href="/get-access">Get Access</Link>
          </DropdownItem>
          <DropdownItem key="transform-property">
            <Link href="/transform-property">Transform a Property</Link>
          </DropdownItem>
        </DropdownMenu>
      </Dropdown>
```

markgoetz:
@nlebovits @thansidwell Is this issue still valid, now that the PR was merged?

bacitracin:
Closing this because I'm pretty sure it's been addressed and is live

PR #324, opened 2024-02-09
fix: increase link hitbox size

markgoetz:
This PR increases the size of the hitbox of the clickable header links by adding the `href` property to the dropdown, resolving #278.

Before:
Screenshot 2024-02-08 at 8 54 58 PM

After:
Screenshot 2024-02-08 at 8 55 14 PM


nlebovits:
@thansidwell approved?

Issue #312, opened 2024-02-05
A11y - Filter/View/Download buttons are not keyboard operable


bacitracin:
## Describe the task
Users should be able to navigate to the buttons via TAB and when they press ENTER, the button activates, same as clicking with a mouse.

## Links
PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/349

Issue #311, opened 2024-02-05
A11y - Add aria-labels to view & download buttons


bacitracin:
## Describe the task
View & download icons have tooltips, but buttons need text associated with them. Add aria-labels on to these two buttons.

https://dequeuniversity.com/rules/axe/4.8/button-name?application=AxeChrome

## Acceptance Criteria

- [ ] When navigating through with Voiceover (CTRL + OPTION + arrow keys) when you land on the view & download buttons, they announce as "View button" or "Download button" instead of just "button".

Note: If you are in Voiceover and TAB through, when you TAB to the button it announces. Some investigation needed to see what's up with that. The buttons aren't keyboard operable, if you TAB to the button and press ENTER, nothing happens. Make sure that the user is landing on the button and not the svg inside, or any other container. 

## Links
PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/349

PR #349, opened 2024-02-22
Bacitracin/311 312 filter btns a11y

bacitracin:
### Description
Addresses the following a11y issues:
- https://github.com/CodeForPhilly/vacant-lots-proj/issues/311
- https://github.com/CodeForPhilly/vacant-lots-proj/issues/312

NextUI's Button component's `onClick` is deprecated.  Switching to `onPress` will capture keypresses in addition to mouse clicks, which will make these buttons keyboard operable.

Added aria-labels for View + Download buttons since they do not include inner button text.

### Testing
- Follow the link below and visit the Map
- You should be able to TAB to Filter / View/ Download buttons, and activate them with ENTER or SPACE keys.
- If you are on the Filter view, the Done button can now be activated with ENTER or SPACE as well.
- If you use a screenreader and navigate to View & Download, they will announce as View button/Download button instead of just "button"


marvieqa:
Looks good and working as expected! 
When we implement the new design, we can use `<aria-labeledby="[tooltip]">`. Currently, it's announced twice as "View Description, View button". 

Let's merge!

marvieqa (approved):
Looks good and working as expected! 
When we implement the new design, we can use `<aria-labeledby="[tooltip]">`. Currently, it's announced twice as "View Description, View button". 

Let's merge!

Issue #329, opened 2024-02-10
A11y: Increase filter info icon size


bacitracin:
## Describe the task
The target size of the info icons on the map filter page are too small. It's currently 16.5px x 16.25px. Minimum is 24x24 px. You can increase the icon itself or the padding.
 
<img width="301" alt="Screenshot 2024-02-10 at 12 22 26 PM" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/7329799/50c4bcf9-9ef0-4095-b08b-bb54e493582c">

## Acceptance Criteria
- [ ] Hoverable area that triggers the tooltip is increased to at least 24 x 24px

PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/350

## Additional context
https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html

Issue #328, opened 2024-02-10
A11y: Make the filter more info icons keyboard navigable


bacitracin:
## Describe the task
If you hover over the little "i" info icons, you get a helpful tooltip. Currently keyboard users and screenreader users are missing out on that info.

Make the info icon focusable/reachable via keyboard nav. When the user TABs to the icon & lands on it, the tooltip should pop up. 
This should fix things for screenreader users to access the information as well, but we'll have to test it to know.

## Acceptance Criteria

- [ ] On the map page open the filters. 
- [ ] Press TAB button. 
- [ ] You should be able to land on the information icons.
- [ ] The tooltip will pop up. 
- [ ] When you TAB away from the icon, tooltip disappears.

## Links
PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/350

marvieqa:
Note: 
Adding to the acceptance criteria:
- [ ] When you press ESC, the tooltip disappears.

PR #350, opened 2024-02-22
bacitracin/328 329 info icon a11y

bacitracin:
### Description
For 
- https://github.com/CodeForPhilly/vacant-lots-proj/issues/328
- https://github.com/CodeForPhilly/vacant-lots-proj/issues/329

Increases the target size of the info icon (just horizontal, didn't want to mess with the vertical alignment and I think the info icons are spaced far enough from the other buttons that it should be ok)

Also makes the info icon keyboard navigable

### Testing
- Follow the link below
- Open up the map filters page (with priority level, parcel type etc)
- You should be able to TAB to the info icons & the tooltip will appear
- You can press ESC or TAB & the tooltip will close


marvieqa (approved):
Working as advertised! ✅

Issue #201, opened 2024-01-17
A11y - Update footer semantics, add aria-label


bacitracin:
## Describe the task
The current footer could use a semantic update & the nav element it uses needs an `aria-label` to differentiate it from the other navs on the page.

## Acceptance Criteria
- [ ] Change the structure of the footer from 
```
<nav>
  <header>
    <ul>
    </ul>
  </header>
</nav>
```

to

```
<footer>
  <nav aria-label="content info">
    <ul>
    </ul>
  </nav>
</footer>
```
- [ ] Give the nav item a descriptive `aria-label` since on some pages there will be up to 3 nav items. TBD on whatever you want ot call it. Currently for screenreader users it just reads as "nav" and will be hard to differentiate.

## Additional context
Example structure (ignore the bit about skip links) https://www.magentaa11y.com/checklist-web/footer/#example-footer

nlebovits:
@all-contributors please add @bacitracin for a11y! :)

PR #351, opened 2024-02-23
fix: Update footer markup structure for accessibility

markgoetz:
This PR updates the footer markup for accessibility per #201.

### DIFF
--- a/src/app/components/Footer.tsx
+++ b/src/app/components/Footer.tsx
@@ -1,31 +1,31 @@
-import { Navbar, NavbarContent, NavbarItem } from "@nextui-org/react";
-
 const Footer = () => (
   <div className="flex flex-col">
-    <Navbar maxWidth="full">
-      <NavbarContent className="flex justify-between items-center w-full">
-        <NavbarItem>
-          <span className="text-sm text-gray-600">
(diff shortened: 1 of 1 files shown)


markgoetz:
The `header` tag is actually hardcoded into NextUI's Navbar component(per [their repo](https://github.com/nextui-org/nextui/blob/main/packages/components/navbar/src/navbar.tsx)), so unfortunately I had to remove it from this component.

bacitracin:
> The `header` tag is actually hardcoded into NextUI's Navbar component(per [their repo](https://github.com/nextui-org/nextui/blob/main/packages/components/navbar/src/navbar.tsx)), so unfortunately I had to remove it from this component.

Ah, actually we talked about this in the last meeting, some of the NextUI components aren't all there a11y-wise, so we want to empower devs to feel free to rewrite components from scratch if the NextUI component has issues.

bacitracin (approved):
👍

bacitracin:
@markgoetz oh wait do you mind updating your base against staging instead of master?

markgoetz:
> @markgoetz oh wait do you mind updating your base against staging instead of main?
Oops, good call!  This is done now.

Issue #337, opened 2024-02-10
A11y: Make filter map buttons actual buttons


bacitracin:
## Describe the task
The filter buttons  (priority level/parcel type/access process etc) are not reachable or operable via keyboard navigation or screenreader. This is because the filters are spans within divs, not actual buttons.

Changing them to buttons will fix this and will also make them visible in Windows High Contrast mode: 
<img width="309" alt="Screenshot 2024-02-10 at 3 11 25 PM" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/7329799/26087299-f950-4e85-8904-b228e4af8346">

## Acceptance Criteria

- [ ] You should be able to TAB to the filter buttons and press ENTER to turn them on or off

## Additional context
* Eventually we will want a new design for these filters. Showing that the filters are on/off via color breaks WCAG rules.
* https://www.smashingmagazine.com/2022/06/guide-windows-high-contrast-mode/
* How to simulate WHCM in Chrome Devtools - https://devtoolstips.org/tips/en/emulate-forced-colors/

marvieqa:
Additional context relates to: #300

marvieqa:
Additional Requirements:

- If we make the filter map buttons actual `<button>`,  use `aria-pressed` attribute so it can communicate the state for screen reader users.
- Or consider marking up these filters as a checkbox instead.

bacitracin:
@marvieqa what do you think of adding a checkmark to the filter to indicate that it's selected?

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/7329799/34bfa8ea-4e01-45d3-8f9b-0148de5fcbc0)

bacitracin:
PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/352

Issue #300, opened 2024-01-30
A11y: Add checkmarks on filter badges when they are selected


marvieqa:
## Description

We should have another indicator that the filter badge is turned on or off. Currently, only color is used to denote whether it is selected or not. Color shouldn't be the only means to convey the meaning of the badges. 

WCAG Criterion: [1.4.1 Use of Color Level A](https://www.w3.org/WAI/WCAG22/quickref/#use-of-color)

I believe the Figma prototype used to have check icons, this but I don't see this anymore.

<details>
<summary> Screenshot </summary>

<img width="1472" alt="Screenshot 2024-01-30 at 17 59 38" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/102739972/4bb4ba8a-eeb7-4dae-8832-e2336b57a120">

</details>

## Acceptance Criteria

- badges should have a symbol or something as additional indication that it is selected.

bacitracin:
PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/352

PR #352, opened 2024-02-25
Bacitracin/300 337/make filter chips btns

bacitracin:
For
https://github.com/CodeForPhilly/vacant-lots-proj/issues/337
https://github.com/CodeForPhilly/vacant-lots-proj/issues/300

Changes filters from Chip component to Button. Uses onPress instead of onClick for keyboard operability.
Adds a Check icon when the filter is active.
Adds aria-pressed for screenreaders.
Makes "Filter" title an h1, and filter headers h2. Tried making the filters themselves h3, but it made things more complicated.


marvieqa (approved):
Good to go! @bacitracin ✅

Issue #331, opened 2024-02-10
A11y: Wrap content on all pages in a <main> tag


bacitracin:
## Describe the task

All content on the page (except skip links) should be contained by landmark tags. This is a best practice.

## Acceptance Criteria

- [ ] Content that is not the header/nav/footer/skip link will be contained in a main tag on all pages

## Additional context
https://dequeuniversity.com/rules/axe/4.8/region?application=AxeChrome

https://dequeuniversity.com/rules/axe/4.8/region?application=AxeChrome

bacitracin:
PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/353

PR #353, opened 2024-02-25
331 / Swap divs to main tag

bacitracin:
For https://github.com/CodeForPhilly/vacant-lots-proj/issues/331

Swaps container div for main tag with id="main". Id will be used for skip link later.

### DIFF
--- a/src/app/components/AboutPage.tsx
+++ b/src/app/components/AboutPage.tsx
@@ -11,7 +11,7 @@ export default function AboutPage() {
   const { isOpen, onOpen, onOpenChange } = useDisclosure();
 
   return (
-    <div className="flex flex-col min-h-screen">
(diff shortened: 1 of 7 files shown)


src/app/components/AboutPage.tsx
```diff
@@ -11,7 +11,7 @@ export default function AboutPage() {
   const { isOpen, onOpen, onOpenChange } = useDisclosure();
 
   return (
-    <div className="flex flex-col min-h-screen">
```
markgoetz, line 14: ❓ Is there a reason we wouldn't make this change in `/src/app/components/Layout.tsx`?

bacitracin: I thought it might affect the skiplink, but it probably won't it's just a bunch of divs. I'll close this PR and try how it sounds.

Issue #304, opened 2024-01-30
A11y: Use regions to divide the "Find Properties" page to provide easier navigation for screen reader users


marvieqa:
## Describe the task

The "Find properties" page can be challenging to navigate by screen reader users because of the number of interactive elements on the page (ex. the map has a ton of clickable area). It would be nice if users have an option to traverse the page without going through every interactive element before they can get to where they want to be. We can mark-up the important sections of the page using `role=region` so screen reader users can scan for this and go straight to the area they wish to be in. 

<details>
<summary> Screenshot </summary>

<img width="1490" alt="Screenshot 2024-01-30 at 22 18 04" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/102739972/696a232e-ce93-4f8c-b566-ec235d536dda">

</details>

## Recommendation

- use `role=region` with an `aria-label` that describes the section
ex. `aria-label="interactive map"`, `aria-label="property results"`

## LInks
PR: https://github.com/CodeForPhilly/vacant-lots-proj/pull/361

PR #361, opened 2024-03-01
304 label map regions

bacitracin:
For https://github.com/CodeForPhilly/vacant-lots-proj/issues/304

### Testing
- Go to the maps page
- Turn on screenreader
- Open landmarks shortcut
- You should see maps/ results sidepanel/controls listed among the other landmarks


marvieqa:
@bacitracin would be nice to have a more descriptive  aria-label like:
- map region: `<aria-label="Interactive map">`
- results regions: `<aria-label="property results">`

Other than this, we are good to ship!

marvieqa (approved):
All the regions are appearing in the VO Rotor. Thanks @bacitracin this is good to go!

Issue #332, opened 2024-02-10
A11y: All links should use blue #0070F0


bacitracin:
## Describe the task
The blue used for the links on the action page #3B82F6 does not meet 4:5:1 color contrast. The blue on the links on the about page pass contrast #0070F0.

Update styles to use this blue throughout.

https://webaim.org/resources/contrastchecker/  https://dequeuniversity.com/rules/axe/4.8/color-contrast

## Acceptance Criteria

- [ ] All blue links will use the darker blue

liathirsh:
Hi, I'd like to take this on!

PR #364, opened 2024-03-02
A11y: All links should use blue #0070F0

liathirsh:
Updated the global.css file to change the link color to  #0070F0 instead of #3B82F6. 
Issue can be found here: https://github.com/CodeForPhilly/vacant-lots-proj/issues/332

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -62,5 +62,5 @@
 }
 
 .link {
-  @apply text-primary underline;
+  @apply text-[#0070F0] underline;
 }


src/app/globals.css
```diff
@@ -62,5 +62,5 @@
 }
 
 .link {
-  @apply text-primary underline;
```
markgoetz, line 65: ❓ Wondering if we should change the value in tailwind.config.css instead - anyone know if there would be any downsides to doing that?

liathirsh: I'm not familiar with Tailwind either, but happy to change Tailwind's config file if that makes more sense to everyone.

markgoetz: I wasn't super familiar with Tailwind either before looking at this project, but they have some pretty good [docs for how to configure colors](https://tailwindcss.com/docs/customizing-colors).

brandonfcohen1: this is the right way to do in tailwind (to the best of my knowledge!)

brandonfcohen1 (approved):
Yes, preference would be to use tailwind

Issue #247, opened 2024-01-22
Fix Additional Map Issues


thansidwell:
- Make all the text in the map legend title case instead of all caps. Priority, High, Medium, and Low.  https://share.cleanshot.com/vFJd8pvj
- Make the text High, Medium, and Low use a darker shade of gray. It's too light for people with vision problems. 
- Remove the "Other" item in the legend. https://share.cleanshot.com/yGgKHVdN
- Add an info icon next to the heading Priority. When a user hovers over a tooltip should appear. The exact text can be added later. https://share.cleanshot.com/jVH0dk35

nlebovits:
flagging that this will probably require us to fork + modify our own copy of https://github.com/markusand/mapboxgl-legend

maybe an opportunity to contribute to that package

Moylena:
@thansidwell  @nlebovits  @brandonfcohen1  Looks like this was deployed, however it's missing the "i" icon - can this be updated and pushed or should this become a new ticket?

thansidwell:
I think adding the "i" icon with the tooltip can be it's own ticket. I'll change the status here to Live.

PR #369, opened 2024-03-05
feat(247): make the mapbox legend more legible

paulhchoi:
## Description
This PR updates the mapbox legend to be capitalized and darker. Due to the way this simply linking to the provided mapbox css, it is admittedly very, very brittle — but given that it's understood that it's a ducttape solution, I'll leave it up to the owners of this project to determine if they'd like to merge it in or not.

## Testing
View the map legend on the main page at `/map`.

#### Before
image

#### After
image

Related to (but technically doesn't close) #247

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -64,3 +64,18 @@
 .link {
   @apply text-primary underline;
 }
+
+/* Styles for the mapbox legend pane. */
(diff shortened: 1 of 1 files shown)


brandonfcohen1 (approved):
I like it, thanks

Issue #376, opened 2024-03-06
BUG: React/JSX errors on /transform-property


brandonfcohen1:
Describe the bug
React/JSX errors on https://www.cleanandgreenphilly.org/transform-property

To Reproduce

Go to https://www.cleanandgreenphilly.org/transform-property
Open Console
Expected behavior
No Errors

brandonfcohen1:
@liathirsh do you want to take this one as well?

liathirsh:
@brandonfcohen1 Sure!

PR #377, opened 2024-03-06
BugFix: React/JSX errors on /transform-property #376

liathirsh:
Issue: https://github.com/CodeForPhilly/vacant-lots-proj/issues/376
It seems like the issue was a p tag that wasn't a parent (changed to a div instead) and a p tag not closed before starting an ul.


brandonfcohen1 (approved):

nlebovits:
@all-contributors please add @liathirsh for code

Issue #310, opened 2024-01-31
On initial load, shows "0 Results" while loading


willonabike:
1. Load the initial /maps page, observe while loading
2. Observed: Page elements start to load
3. After a short time
Expected: Informs user that results in progress e.g. “Loading Results”
Actual: Shows “0 Properties in view”
4. After full page load, correctly updates to the “15k Properties in view”

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1652510/7a65284f-55a3-46e5-821d-d672ed1b2b70)

jgaye:
@brandonfcohen1 @willonabike @thansidwell  
When you have time, my PR is waiting for review [here](https://github.com/CodeForPhilly/vacant-lots-proj/pull/379)  

It's very nice how you have a branch App auto-built for testing with Vercel :+1:

thansidwell:
This looks great!

PR #379, opened 2024-03-08
jgaye/310 fix side panel loading state

jgaye:
# Related issue
Solving issue #310 

# Changes
1. Display a text and spinner in the PropertyDetailSection while properties are being loaded
2. Hide the SidePanelControlBar while properties are being loaded

Design based on Slack conversation [here](https://codeforphilly.slack.com/archives/C05H9QBMP96/p1709839246741599).

# How to test
Start the server (with backend data) and go to the [localhost:3000/map](localhost:3000/map) page. While the map and properties are being loaded from the backend, the text and spinner will be displayed on the right hand side.  
The text and spinner should be replaced by the buttons of the SidePanelControlBar and the content of the PropertyDetailSection as soon as the data is available.  
NB: if you use the React Developer Tools, you can toggle the `loading` prop to `true` in the Components to re-display the loading state, as shown in the screenshot attached
image

### DIFF
--- a/src/app/components/PropertyDetailSection.tsx
+++ b/src/app/components/PropertyDetailSection.tsx
@@ -61,7 +61,7 @@ const PropertyDetailSection: FC<PropertyDetailSectionProps> = ({
   }, [page, featuresInView]);
 
   return loading ? (
-    <Spinner />
+    <Spinner className="flex flex-wrap" />
   ) : selectedProperty ? (
     <SinglePropertyDetail
       property={selectedProperty}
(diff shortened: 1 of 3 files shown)


brandonfcohen1:
This PR looks great for the loading, but I think we also want to show something better than '0 Properties' when you're zoomed out, right?

@thansidwell should we create a new ticket for that?

thansidwell:
@brandonfcohen1 I think zooming out the and properties disappearing is a disorienting experience. And it's best solved by...

1. Somehow making properties visible even when zoomed out. (Ideal)
2. Or NOT allowing people to zoom out to the point where the properties disappear.

So yeah, I'd advocate for a ticket with either of those options.

nlebovits:
@all-contributors please add @jgaye for code

Issue #299, opened 2024-01-30
A11y: Property card component a11y improvements


marvieqa:
## Description

The cards that represent each property are not keyboard-navigable. It could be made more accessible.
Consider using this card component as a pattern:
https://inclusive-components.design/cards/

## Recommendation

<img width="899" alt="Screenshot 2024-01-30 at 16 19 36" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/102739972/ec039d58-ad0b-4178-a397-7313bfec92e9">

- The whole card is clickable by mouse
- Only the card title is marked up as a link to preserve its semantic meaning for assistive technology users.
- Only the card title receives keyboard focus when keyboard-navigating.
- Card image is decorative

WCAG Criteria:
[2.1.1 Keyboard Level A](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html)
[1.3.1 Info and Relationships Level A](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html)

PR LINK - https://github.com/CodeForPhilly/vacant-lots-proj/pull/383

markgoetz:
I'd love to make the clickable area being a link, but it's going to need a bit of rearchitecture so that each property can have its own page (cc @brandonfcohen1).  Is it worth making it a button in the short-term?

PR #383, opened 2024-03-09
Bacitracin/299 prop card a11y

bacitracin:
For https://github.com/CodeForPhilly/vacant-lots-proj/issues/299

Some updates to make the property cards in the list keyboard navigable.
- I went with making the address a button instead of a link because the url doesn't currently change when you click to see the single property detail.
- Made card image decorative (left the description on single property view, TBD on whether we want to keep that or not)
- The whole card is clickable


bacitracin:
@marvieqa finally made this PR deployment happy, it's ready for a11y testing

brandonfcohen1 (approved):

Issue #261, opened 2024-01-23
Responsive: Fix padding and width with content pages


thansidwell:
Right now the text on content pages...

- Goes very wide, past the standard recommended width for ideal reading. https://share.cleanshot.com/hb90DnPD
- On mobile size screen there's no padding. https://share.cleanshot.com/gcQ05CQP

Make the fixes https://share.cleanshot.com/FqW84lqb

- add a 1000px max-width to the 
- Add a 20px padding on mobile size screens

bacitracin:
@jgaye iIsaw that there was a PR merge, but also a revert. What's the status on this ticket - still in development or live?

jgaye:
@bacitracin it's still in development. I haven't had the chance to redo the changes since the directory architecture changes, I hope to do as new PR tomorrow

jgaye:
> [@bacitracin](https://github.com/bacitracin) it's still in development. I haven't had the chance to redo the changes since the directory architecture changes, I hope to do as new PR tomorrow

New PR in out @bacitracin

thansidwell:
@jgaye I see it's been done to the Get Access and Transform pages. Could you also do it to these pages? 
- Home
- About

jgaye:
> > @jgaye I see it's been done to the Get Access and Transform pages. Could you also do it to these pages?
> > 
> > * Home
> > * About
> 
> @thansidwell because Home and About are on two columns, I'm not sure it's worth adapting with the same max-width, as each column is already smaller than half the size of the screen. Additionally, the column format also gives the mobile padding by default.
> 
> How would you apply the max-width for those pages?

@thansidwell Asa very quick demo, this is how the About page would look like with the max-width applied to the whole page:  
![image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/4049458/f326ef3a-1099-4605-9c4c-321f356555b0)

jgaye:
@thansidwell sorry for all the messages.
I updated the PR with a max width for About and Home. that is the same as the other pages, and adapted the columns so that there is no double padding applied. Please have a look and let me know if that's good.

thansidwell:
@jgaye With some other tickets we're removing the columns with the new content coming in.

thansidwell:
I approved the pull request.

Moylena:
Should this be considered closed? I see in the PR that there is 1 pending review? Not sure if that leaves this in limbo

thansidwell:
[...]

PR #402, opened 2024-03-12
Jgaye/261 handle sizing of take action pages

jgaye:
For #261 

@thansidwell I applied the max-width and margin for the 3 pages under 'Take action' section.
I did not touch the 'Overview' and 'Methodology' page has they have 2 columns and already have a padding on mobile view.

I'm not sure if it would be a better practice to create a `<TakeActionContent>` Component to reuse for those pages, instead of adding it for each page. Not sure who to tag that would have an opinion on that. Can you tag someone @thansidwell ?

### How to reproduce
Got to /transform-property, /take-action-overview or /get-access, and manipulate the screen to see the padding for wide screen, or mobile size screen

### DIFF
--- a/src/app/components/GetAccessPage.tsx
+++ b/src/app/components/GetAccessPage.tsx
@@ -187,76 +187,76 @@ export default function GetAccessPage() {
           [https://groundedinphilly.org/#pathways], which can better help you
           understand if you qualify.
         </p>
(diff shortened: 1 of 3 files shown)


nlebovits:
Thanks @jgaye. Maybe @paulhchoi has opinions on best practices here?

brandonfcohen1 (approved):

Issue #259, opened 2024-01-23
Responsive: Make Navigation Responsive


thansidwell:
Right now on small devices, the navigation becomes hidden. https://share.cleanshot.com/xvPmddB0

Make these changes to make the app responsive and accessible on mobile devices. 

- Make the navigation not hide on mobile. https://share.cleanshot.com/BvXjnLDy
- When the viewport reaches a smaller size and can no longer fit showing all the navigation option, somewhere around 600px https://share.cleanshot.com/TN3sXRmq
- Display a button saying "Menu" with a three-line menu icon 
- When a user clicks on the "Menu" icon a dropdown should appear with the navigation options in it. They should also include the icons. https://share.cleanshot.com/x9m2L0M8
- Clicking on these navigation items should navigate user to the different pages. 

Here's a prototype you can reference. https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=203%3A14992&mode=design&t=BCw3uF1EeDQPocXH-1

Moylena:
UPDATED DESIGN: https://www.figma.com/proto/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?page-id=187%3A12602&type=design&node-id=2592-30019&viewport=-657%2C-623%2C0.1&t=fqZvOvLyE9qv7AAV-8&scaling=min-zoom&starting-point-node-id=2592%3A30019&hide-ui=1

thansidwell:
Like @Moylena noted. The options in the navigation have changed but the experience hasn't changed.

Amberroseweeks:
I'd like to assign this to myself, if this task is open. I worked on the desktop view. I just wonder what tablet view might look like and if there are any specific breakpoints we would like to follow?

Moylena:
@thansidwell  Do we have specific breakpoints for mobile, tablet, and (related to another ticket on max width) desktop?

thansidwell:
@Moylena Sorry! I should have specific break points, but I haven't laid them out. It seems that @Amberroseweeks used her own judgment and it turned out fine.

PR #418, opened 2024-03-13
259-responsive-nav-feat-css-nav-item-cleanup

Amberroseweeks:
### Mobile Navigation Added

I used Next UI's NavbarMenuToggle component to add the mobile navigation to the site. I tried to follow the design closely and used many tailwind classes to edit the default component to replicate it. The mobile breakpoint in 640. We should consider a new issue for tablet, which would be between 768px and 1024px I believe.

### There is room for improvement for a11y 
The DOM order is not accurate. For some reason the menu list is rendering at the end of the page and making the links appear at the end of the tab order when they should appear as the 4th, 5th, and 6th option after selecting the menu to open. This is only on mobile so it's not a huge issue but definitely needs a new issue to correct this as responsive design helps users who zoom in at 300% or more on desktops as well. 

### CSS Updates to nav items
I updated some CSS selectors and adjusted some styling to better match the Figma design 

This closes issue #259 (https://github.com/CodeForPhilly/vacant-lots-proj/issues/259)

### DIFF
--- a/src/app/components/Header.tsx
+++ b/src/app/components/Header.tsx
@@ -1,17 +1,8 @@
-import {
-  Navbar,
-  NavbarContent,
-  NavbarBrand,
-  NavbarItem,
-  Dropdown,
-  DropdownTrigger,
-  DropdownMenu,
-  DropdownItem,
-  Button,
-} from "@nextui-org/react";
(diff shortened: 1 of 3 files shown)


marvieqa (approved):
menu works in mobile. We'll create a new ticket to address accessibility issues in mobile.

PR #417, opened 2024-03-13
feat(338): add new pitchdeck slides to homepage

paulhchoi:
## Description

This PR updates the homepage based on the latest figma guidelines ([linked here](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2266%3A23177&mode=design&t=jX4y1auMbn2sVoGD-1)).

Leaving as draft for @thansidwell and stakeholders to review — specific feedback requests include:
1. I copied some of the graphics directly from the figma, but I believe you meant some of them as placeholders for now — would love your feedback on the finalized versions of those section graphics?
2. I assumed the final "Get Started" link went to `/map`, but that was a shot in the dark.
3. I could use more direction on the mobile view sizes for the "three blue boxes" graphic.

## Desktop (1980px)
screencapture-localhost-3000-2024-03-13-01_11_30

## Mobile (375px)
screencapture-localhost-3000-2024-03-13-01_11_48

Closes #338

PR #435, opened 2024-03-15
feat(434): refactor image component usage

paulhchoi:
## Description

This PR updates the usages of the `Image` component from `nextui` to `next/image` on the homepage, /about, and /methodology pages to speedup pageload times and increase our lighthouse score.

Note, leaving as draft until https://github.com/CodeForPhilly/vacant-lots-proj/pull/417 is merged into staging as this PR's been branched off of that (to include changes made to the homepage)

## Demo

**homepage page, before**
434-homepage-01-before

**homepage page, after**
434-homepage-02-after

**`/about` page, before**
434-about-01-before

**`/about` page, after**
434-about-02-after

Closes #434

PR #436, opened 2024-03-15
feat(421): dedupe `<title />`s by organizing layout and page files

### LINKED ISSUE #421: Multiple Page Titles Tags on Interior pages
**Describe the bug**
The interior pages of the website have multiple page titles. This is confusing for accessibility reasons and probably not very good for SEO. I added the fix for unique page titles, but adding in the <title> tag with the title is creating a new issue.

**To Reproduce**
1. Inspect an interior page such as Get Access, view the < head >
2. See that there will be three < title > tags

**Expected behavior**
Only one unique title tag per page.

**Screenshots**
Screen Shot 2024-03-14 at 9 36 42 AM

**Potential Fix**
I avoided using the Next Metadata API for this initially because our page components are not server components and next metadata only works on server components. (reference: https://stackoverflow.com/questions/76445050/not-able-to-change-the-title-with-metadata-when-using-use-client-in-next-js) So I believe a good fix could be converting the page.tsx to a server component, moving the interactive elements to a new component then importing those elements while adding the metadata information in each specific page.tsx file. This is a good tutorial: https://nextjs.org/learn/dashboard-app/adding-metadata

I am not sure if there are any implications to making this change though. I tested it out on a branch and it's working so far. 
Screen Shot 2024-03-14 at 9 40 30 AM

paulhchoi:
## Description

This PR's diffs seem much more drastic than they actually are, as its more about re-organizing code to simplify the components folder more than anything:
- I moved the `src/app/components` dir one level up to `src/components` to make the `src/app` dir just for routing files and page components, and `src/components` just for actual shared components.
- I removed the `<Layout />` component and added its contents to the root `layout.tsx`, so it'd automatically wrap all nested pages, without needing to import it anywhere else. (This root `layout.tsx` also exports the 'base' metadata object along with a `title.default`/`title.template` so nested layouts can simply extend it)
- I added page-level `layout.tsx`'s for all pages (eg: `app/about/layout.tsx`), and extended the title templates defined in the root metadata object.
- I also move the `{name}Page.tsx` components to sit in each page directory — this again simplifies the `component` dir to just be shared components, and moves the specific, only-used-in-one-page components can be nested with its page directory.
- I also installed and used `react-icons/pi` for the phosphor icons to use icons as simple server-side rendered svg's, as the `@phosphor-icons/react` needed to be wrapped in a `<NextUIProvider />`, forcing the page its used in to be client-side rendered — which seems silly for a simple svg.

## Testing (look for `<title>` on bottom of images`)

User should only ever see one `<title />` element in the head of any page.

**Before**
image

**After**
image

I branched this off of https://github.com/CodeForPhilly/vacant-lots-proj/pull/417 again, so keeping this as draft until that gets merged in and to also get feedback.

Closes #421


paulhchoi:
@brandonfcohen1 would love your feedback on these changes, specifically regarding the reorg of the files and folders, as its pretty sweeping.

I'll also leave the conflicts until both https://github.com/CodeForPhilly/vacant-lots-proj/pull/417 and https://github.com/CodeForPhilly/vacant-lots-proj/pull/435 are merged, as I expect some more to crop up before pushing this one out.

cc: @Amberroseweeks

nlebovits:
@paulhchoi thank you. you're a hero. <3

paulhchoi:
@nlebovits @brandonfcohen1 ready for review

Issue #260, opened 2024-01-23
Responsive: make Find page Mobile Friendly


thansidwell:
The Find page needs to work on mobile devices.

- Add a button that toggles between the map and list views of properties
- Make the property card list appear in one column
- When someone clicks on a property on the map it should switch to the property detail view

Here's a link the the prototype in Figma. https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2421%3A22712&mode=design&t=EYanZQnoOdgWCTls-1

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/49b1f541-5c25-4699-b826-dc1ada7a8dfc)

Moylena:
NOTE design is updated but doesn't impact the requested changes

amyyeung17:
Hello! I would like to be assigned to this issue if it's available. :)

PR #442, opened 2024-03-18
Map & properties are responsive and can be toggled in small screens

amyyeung17:
## Description
This PR addresses issue #260 

The find page is now responsive and can toggle between the map/properties in small screens (< 640px, following Tailwind's sm breakpoint). The toggling is handled by a state and display: none with Tailwind variants. 

I made few styling changes and added some code to handle side effects. Namely, rendering the properties separately affected the pagination's styling on the initial load. 

### Desktop to mobile

https://github.com/CodeForPhilly/vacant-lots-proj/assets/28910543/1df3d7b3-9106-4ba7-b9ab-2a4a12ed1f16

### Mobile interactions 

https://github.com/CodeForPhilly/vacant-lots-proj/assets/28910543/8295bc55-5157-43e6-ab44-6e761aa6c269

### Overflow for x & y 

https://github.com/CodeForPhilly/vacant-lots-proj/assets/28910543/bf5eabdd-40c0-40b5-b187-844b2a4d8ede


amyyeung17:
Hi @thansidwell @Moylena! I was hoping to get some clarifications. 

How should the tooltips and filter work for small screens? For the latter, does this mean pressing on filter would swap to the properties component? 

A big problem has been the control bar. It overflows for almost all screens < lg-ish and the options aren't visible. This also causes double-scrolling (x & y), which I was able to fix with the exception of a small interval around 750px. At this point, the pagination bar also overflows. I'm not sure if we wanted to handle that here or in a future PR. 

Thanks! :)

thansidwell:
@amyyeung17 Great work! Here's a video and a list of the issues I noticed. https://share.cleanshot.com/lCsH0zGY

1. Change the View toggle button to toggle between a map and a grid of four squares to toggle between the map and the cards. I realized we don't have the toggle function adding to the button component, and I think this is approach is better. The icons are attached here. 
GlobeHemisphereWest
SquaresFour
2. Stop the content from shifting up and down. If you can. 
3. When you are looking at the map, click on a property, and the property shows up, when you click Back you should go back to the map. 
4. You should be able to click on the filter on the map view. Right now it doesn't work. I'm guessing that's just tricky because of where it's located in the HTML.

amyyeung17:
@thansidwell Noted, thanks for the feedback! I think I can do 2 now that the navbar is merged. I haven't thought much about the rest, but I don't foresee too many difficulties.

amyyeung17:
@thansidwell Addressed your comments and took a shot at making the info less cramped. As expected, 2 was resolved with the navbar. :) I also added some hover states for some buttons and property cards. 

There are still several things to work on like the nav options, pagination, and odd scrolling behavior. Perhaps these would be best done in new PRs? 

Here's the updated version: 


https://github.com/CodeForPhilly/vacant-lots-proj/assets/28910543/bc0a994e-8906-4891-b82b-1094986cd016

amyyeung17:
@thansidwell Realized when working on a different PR that the filter option acted funky if it's applied in-between resizing and added some fixed. Besides that, I think this PR is done.

https://github.com/CodeForPhilly/vacant-lots-proj/assets/28910543/6bd9163a-813a-4cf0-b925-5d236d90bb1e

thansidwell:
Hey @amyyeung17! Sorry, I didn't respond earlier. Somehow I missed this. It looks great! 

There are a few minor issues that are being resolved with other tickets or I think are best made their own ticket.

thansidwell (approved):
It all looks good to me. 

There are some minor issues we are resolving or will resolve with other tickets.

Issue #413, opened 2024-03-13
Small: Update # of properties to show sum total number with commas


Moylena:
Currently map returns # of properties as (for ex) 2000

Update to reflect the values with a comma (for ex) 2,000

paulhchoi:
sorry this should be in 'Staging', not 'Live', that was my mistake closing this ticket!

thansidwell:
@jgaye Looks good in the pull request preview!

nlebovits:
closed by #453

PR #453, opened 2024-03-20
413 - Format number into string with comma

jgaye:
For ticket #413 

# How to check  
Go to the /map page, and zoom out to have more than 1,000 properties. The number should be formatted with a comma between thousands and hundreds.  
Note that the formatting should adapt to the Locale of the browser, meaning that it can change if you are on a VPN to adapt to the country's way of writing numbers (for example 1.500 in France)

@Moylena


paulhchoi (approved):
LGTM 👍

PR #467, opened 2024-03-23
fix(448): fix overscrolling on the map page

paulhchoi:
## Description

This PR fixes the over-scrolling on the `/map` page. This is admittedly not the best fix, but its a patch that'll work for now given our usage.

## Demo

https://github.com/CodeForPhilly/vacant-lots-proj/assets/8061917/b44be189-51cb-4297-b76e-1809de44d99c


thansidwell (changes requested):
@paulhchoi The footer is fixed taking up valuable space on the screen all the time. I think it would be best to just remove it on the Map page. Or if we feel we have to have to have it only appear on the right panel side under the list of properties and not fixed. Does that make sense?

paulhchoi:
> @paulhchoi The footer is fixed taking up valuable space on the screen all the time. I think it would be best to just remove it on the Map page. Or if we feel we have to have to have it only appear on the right panel side under the list of properties and not fixed. Does that make sense?

@thansidwell good catch, I missed that last bullet in the acceptance criteria! last fix shifts the footer into the bottom of the sidebar, and added a flex-wrap so it nests into a column on smaller viewports. please test again.

thansidwell:
@paulhchoi It all looks good! I don't know how to approve it now...

paulhchoi:
> @paulhchoi It all looks good! I don't know how to approve it now...

@thansidwell just clicked the 'request re-review' so you should be able to now?

brandonfcohen1 (approved):

Issue #399, opened 2024-03-12
A11y: When Google Street View is open, focus does not move to the street view


bacitracin:
## Describe the task
When opening Google Street view for a single property, the focus is still on the single property sidepanel where you triggered the view from. 

Here's what the experience is like for screenreaders:

https://www.loom.com/share/605fba1977af458293337021e074ed47?sid=998d512e-f4a3-4d6f-8cf0-c08847a260c7

## Acceptance Criteria

- [ ] When opening the full screen Google street view, focus will move from the trigger button to the street view container.

Note: this issue is just for moving focus, not for fully working keyboard/screenreader use within the street view

Amberroseweeks:
Hi, i'd like to assign myself this task if it is still available. Thank you!

Amberroseweeks:
Just sharing an update, I have focus applied to the streetview but now that I am in here I am thinking we should add a couple more updates like:
-Close iFrame on pressing ESC key
-Have Close button receive focus first 

So I am working on having those items included. @bacitracin @marvieqa let me know if there is anything else you think should be added

PR #483, opened 2024-04-01
Amberroseweeks/399 google street view focus trap

Amberroseweeks:
This closes issue: https://github.com/CodeForPhilly/vacant-lots-proj/issues/399

**Changes:**

- User's focus using the tab key is trapped inside the iframe
- Close button is first to receive focus when iframe is open
- focus is returned to trigger element when iframe is closed

**Consider adding in the future**

- Close iframe on ESC key press

I struggled with this a little, I had to refactor the close button and add new logic to the StreetViewModal to create a focus trap and allow for the close button to be within tab order. It's actually not in tab order now. It is focusable but difficult to find.

I tried to implement the ESC key as another way to close the full page iframe but when you interact with the iframe inside, the ESC key function will not fire. If you don't focus on the iframe the ESC will work. 

When the iframe is closed, the focus will return to the trigger element the button with the id "outside-iframe-element"

### DIFF
--- a/src/app/map/page.tsx
+++ b/src/app/map/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { useEffect } from "react";
+import React, { useEffect, useRef } from "react";
 import {
   FilterView,
   Footer,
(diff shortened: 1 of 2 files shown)


bacitracin (approved):
Much nicer experience, great work!

PR #485, opened 2024-04-02
460 Correct the Find Properties Tab Title and URL

AZBL:
This change changes the title of the tab for the Find Properties (map) page from Map to Find Properties and the url from /map to /find-properties.

It involved changing the title from Map to Find Properties in layout.tsx as well as renaming the map directory to find-properties. This in turn led to numerous files be changed due to imports. 

Some changes were automatically made by Prettier as well.


AZBL:
@paulhchoi @thansidwell @brandonfcohen1

thansidwell (approved):
Nice!

PR #499, opened 2024-04-09
Add metadata to Streetview images, allow deep linking

brandonfcohen1:
I added metadata to the streetview objects in GCP which allows us to essentially use it as a database of sorts. The metadata has location and capture date. You can access it by `OPA_ID` like this:

`https://storage.googleapis.com/storage/v1/b/cleanandgreenphl/o/${opa_id}.jpg`.

I used this to set the URL to change when in a property detail view and allow for deep linking. @thansidwell I'm sure you'll want to add a Share button, will let you mock that up.

I also cleaned up the streetview script so it can run in Docker.

https://www.loom.com/share/9d231645443b476e9aa2568004f38072

### DIFF
--- a/src/app/find-properties/[[...opa_id]]/page.tsx
+++ b/src/app/find-properties/[[...opa_id]]/page.tsx
@@ -1,26 +1,33 @@
 "use client";
 
-import React, { FC, useRef, useState, useEffect } from "react";
+import React, { useRef, useState, useEffect } from "react";
 import {
   FilterView,
-  Footer,
   PropertyDetailSection,
   PropertyMap,
   SidePanel,
   SidePanelControlBar,
 } from "@/components";
 import { FilterProvider } from "@/context/FilterContext";
-import { NextUIProvider } from "@nextui-org/react";
+import { NextUIProvider, Spinner } from "@nextui-org/react";
 import { X } from "@phosphor-icons/react";
 import { MapGeoJSONFeature } from "maplibre-gl";
-import StreetView from "../../components/StreetView";
+import StreetView from "../../../components/StreetView";
 import { centroid } from "@turf/centroid";
 import { Position } from "geojson";
-import { ThemeButton } from "../../components/ThemeButton";
+import { ThemeButton } from "../../../components/ThemeButton";
+import { useRouter } from "next/navigation";
+import { ViewState } from "react-map-gl";
 
 export type BarClickOptions = "filter" | "download" | "detail" | "list";
 
-const MapPage: FC = () => {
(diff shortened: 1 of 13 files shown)


thansidwell (approved):
Looks good to me!

src/app/find-properties/[[...opa_id]]/page.tsx
```diff
@@ -29,11 +36,86 @@ const MapPage: FC = () => {
...
+  const linkedPropertyRef = useRef<string | null>(null);
+  const [initialViewState, setInitialViewState] = useState<ViewState>({
+    longitude: -75.15975924194129,
+    latitude: 39.9910071520824,
+    zoom: 13,
+    bearing: 0,
+    pitch: 0,
+    padding: { top: 0, bottom: 0, left: 0, right: 0 },
+  });
+  const [linkedPropertyParsed, setLinkedPropertyParsed] = useState(false);
```
Comment (paulhchoi) [inline src/app/find-properties/[[...opa_id]]/page.tsx:54]: this is 1000% a nit lol but i prefer to have boolean states start with `is`, so would be nice to have this as `isLinkedPropertyParsed`

paulhchoi (approved):
fantastic work @brandonfcohen1, looks great to me!

paulhchoi:
oh we may also want to consider the initial screen shown when a user deeplinks on mobile — whether we want to show a zoomed in map or the single property detail pane

src/app/find-properties/[[...opa_id]]/page.tsx
```diff
@@ -29,11 +36,86 @@ const MapPage: FC = () => {
...
+    let { location } = metadata;
+    location = JSON.parse(location.replace(/'/g, '"'));
+    setInitialViewState({
+      ...initialViewState,
+      longitude: location.lng,
+      latitude: location.lat,
+      zoom: 18,
+    });
+    setLinkedPropertyParsed(true);
+  };
```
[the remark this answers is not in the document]
Comment (brandonfcohen1) [reply src/app/find-properties/[[...opa_id]]/page.tsx]: ah thanks, I intended to catch that but didn't test for if it passed the regex but wasn't valid. just pushed up a fix

src/app/find-properties/[[...opa_id]]/page.tsx
Comment (brandonfcohen1) [reply src/app/find-properties/[[...opa_id]]/page.tsx:54]: fixed :)

brandonfcohen1:
> oh we may also want to consider the initial screen shown when a user deeplinks on mobile — whether we want to show a zoomed in map or the single property detail pane

yeah at first I did that and had the map pan but made it slower- @thansidwell what do you think? will close this issue and you can make a new one

Issue #508, opened 2024-04-11
Feature: Filter by Neighborhood and Community Organizations


thansidwell:
Add the ability to fitler by these two fields matching the [prototype](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2592%3A26981&mode=design&t=TMx1kZ7ygrhbLR1X-1). 

- Neighborhoods
- Community Organizations

I believe the backend work has already been done and this data is in the database. 

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/12b266da-fe8e-44a1-a0a8-b7803fcfa4ca)

This should use a multi-select field component with all the appropriate state matching the [design](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=3386%3A6293&mode=design&t=TMx1kZ7ygrhbLR1X-1). 

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/18e13d23-6367-4a24-bd87-7ad17416fb5e)

jgrimes86:
Hello @Moylena.  I would like to try to resolve this issue.  Can I be assigned to it?

thansidwell:
Thanks, @jgrimes86! I assigned you.

nlebovits:
Thanks @jgrimes86! @paulhchoi if you can review Jim's PR when ready, that would be great

brandonfcohen1:
Per convo with @jgrimes86 on slack, we'll need to display RCO better in the database, right now we just mash a bunch of contact info together. I'll make a new ticket for that

brandonfcohen1:
https://github.com/CodeForPhilly/vacant-lots-proj/issues/511

brandonfcohen1:
@jgrimes86 I added a field called `rco_names` which has names you can filter on for the RCOs. You can add to https://github.com/CodeForPhilly/vacant-lots-proj/pull/512 or create a new PR

jgrimes86:
Thank you @brandonfcohen1.  I'm working on some requested changes to https://github.com/CodeForPhilly/vacant-lots-proj/pull/512, so I will add it there.

nlebovits:
closed by #511, #512

PR #512, opened 2024-04-14
#508: Filter by Neighborhood

jgrimes86:
This change adds a feature to filter properties by neighborhood.  This feature can also be used to filter by community organization, pending some changes to how RCO data is stored in the database (see issue 511).

This new filter includes the following features:
- A list of neighborhoods is rendered in a NextUI Select component.  Multiple neighborhoods can be selected from the dropdown menu.
- When a neighborhood is selected from the list, a NextUI Chip component for the neighborhood is rendered in the Select field, and the selected neighborhood is removed from the list of neighborhoods in the dropdown menu.
- Neighborhoods can be removed from the filter by clicking the 'x' on the neighborhood's Chip component.  After all the selected neighborhoods are removed from the filter, the map returns to showing all properties within the field of view.

This addresses issue #508


jgrimes86:
Hello @paulhchoi.  Can you review this PR?

paulhchoi:
Hey @jgrimes86, I won't have access to a laptop til later tonight but will take a look as soon as I can 👍

brandonfcohen1:
possible this was how we wanted it, but I don't love how the neighborhoods disappear from the dropdown when selected. thoughts @thansidwell ?

thansidwell:
@jgrimes86 @brandonfcohen1 Some details don't quite match the prototype. 

CleanShot 2024-04-14 at 18 30 05@2x

Here's the changes. 
- The icon in the pills should be an X icon, not with a circle around it and filled. Ideally from the Phosphor icon library we've been using throughout the site. 
- The padding on elementsneeds to get adjusted. While you can't get Dev mode you should be able to see property details as Amber notes here. https://codeforphilly.slack.com/archives/C05H9QBMP96/p1710189823519639?thread_ts=1710185385.458249&cid=C05H9QBMP96. If not here's some screenshots of details. 

CleanShot 2024-04-14 at 18 33 27@2x

CleanShot 2024-04-14 at 18 33 03@2x

CleanShot 2024-04-14 at 18 32 38@2x

CleanShot 2024-04-14 at 18 34 00@2x

thansidwell:
@brandonfcohen1 And to your concern about the options disappearing from the dropdown when selected. I've seen well-designed systems do it both ways. And I think both ways can be very usable. I think we can keep it until we find more clear evidence to lead us another way.

src/components/Filters/DimensionFilter.tsx
```diff
@@ -25,15 +27,91 @@ const DimensionFilter: FC<DimensionFilterProps> = ({
...
+    //   return rcos_detail[rco]
+    // })
+    dispatch({
+      type: "SET_DIMENSIONS",
+      property,
+      dimensions: newMultiSelect,
+    });
+  }
+
+  if (property === "neighborhood" || property === "rco_info") {
```
paulhchoi, line 54: Instead of hardcoding the property types here, it may be more flexible to add a new `type` field of either `multiSelect` or `buttonGroup` or something to the `filter` object, and switch off of that instead. Also, given the duplication of the tooltip and header code on both of these, might be worth DRYing that up and creating new filter components to pass into

paulhchoi (changes requested):
appreciate your work on this @jgrimes86! left some notes, would love to hear your thoughts

src/components/Filters/FilterOptions.tsx
```diff
@@ -0,0 +1,518 @@
+
+export const neighborhoods: string[] = [
```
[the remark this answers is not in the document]
jgrimes86: Definitely a good idea.  I hadn't thought about narrowing down the list based on the map view.  I'll work on that this afternoon.

src/components/Filters/DimensionFilter.tsx
jgrimes86: Makes sense.  I'll work on revising those sections.

jgrimes86:
@paulhchoi, I am close to getting the dynamic rendering of neighborhood list working, but there is a quirk in maplibre's queryRenderedFeatures function that I wanted to point out so that it doesn't become a hidden issue.

As you zoom out in the map view, getting features from the map with `const features = map.queryRenderedFeatures(bbox, { layers })` (pre-existing code on line 206 of PropertyMap.tsx) does not return a feature for every property in the database.  After a certain point, the total number of features retrieved decreases, so that when the entire city is in view there are only about 7600 features being returned. 

On the one hand, that seems to be enough of a sampling to make a list of all the neighborhoods with features/propertiess in them.   Zoomed all the way out, I am able to get a list of 151 neighborhoods from features, which is the total number of neighborhoods with properties in our database. (There are 158 "neighborhoods" in the OpenPhillyData database, but seven of them ('Crestmont Farms', 'East Park', 'Mechanicsville', 'Pennypack Park', 'West Torresdale', 'Wissahickon Hills', 'Wissahickon Park') don't have any properties in our database, so they can't be included in a list made from our database properties.)

On the other hand, I don't know enough about statistics or maplibre to understand whether a set of ~7600 properties is enough to guarantee that every neighborhood will be represented every time. 

To demonstrate the numbers we're dealing with: starting at the initial view and zooming out results in a varying and decreasing number of features returned from the map.  Here is a console log of the feature returned with three clicks of the 'zoom out' button on the map:
image

Checking the number of neighborhoods in view with the same changes in zoom level ends up with 151 neighborhoods
image

How should I proceed?

brandonfcohen1:
Hey @jgrimes86 - that's correct. At lower zoom levels the tiles drop overlapping features for performance reasons. Issues like this are the downsides of not having a proper database :)

The neighboorhood list should ~never change, so I'm fine with hardcoding the values in a JSON files somewhere. Same with RCO data. Thoughts @paulhchoi ?

jgrimes86:
@paulhchoi and @brandonfcohen1, I was overoptimistic yesterday when I thought I was close to having the dynamic neighborhood filter working.  I can get the dynamic list to render in the multi-select menu, but when an option is selected the array of options gets changed to `[undefined]`, breaking the multi-select.  I've tried different approaches but can't get it working.  Using the hard-coded lists of neighborhoods and rcos may be necessary, regardless of our preferences (at least until someone besides me can figure out a solution).

paulhchoi (approved):
@jgrimes86 thanks for the extra work put into this! based off your extra work and brandon's note, lets revert back to your hardcoded implementation, with plans to look into the dynamic list in the future. look great otherwise, fantastic job on this!

thansidwell:
@jgrimes86 Just wanted to jump in that the multi-select component looks great! 

CleanShot 2024-04-16 at 10 12 44@2x

jgrimes86:
@paulhchoi I made one more commit with minor code formatting changes and removal of the large, unused `rco_details` object from FilterOptions

brandonfcohen1:
this looks good, not sure why git is showing python changes, I'll figure it out then merge

brandonfcohen1 (changes requested):
@jgrimes86 can you merge in the latest staging? I'm not sure why git is showing diffs here... really weird.

Alternatively, if you can give me push permissions to your fork I can play around with it

jgrimes86:
@brandonfcohen1 I think I successfully merged the latest staging into my branch just now.  I'll give you permissions to my fork in case something needs to be fixed

brandonfcohen1:
> @brandonfcohen1 I think I successfully merged the latest staging into my branch just now. I'll give you permissions to my fork in case something needs to be fixed

that worked, thanks! will merge- nice work here!

thansidwell:
Hey @jgrimes86! The Neighborhood filter works great, but when you filter by Community group turns up 0 properties every time. That was part of the ticket #508. 

https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/0910dcb2-a4e9-447e-86ea-ce10064f0104

I wrote a new ticket #550 to make it happen.

Issue #430, opened 2024-03-15
Content + Styling: Create content cards for "Transform a Property" - Green


Moylena:
Create new GREEN content + image cards under the headers:

Basic interventions
Advanced interventions

This ticket does NOT include the BLUE content cards

Note headers within the cards are final, however you can greek the remaining content that is not fully baked yet. 

See Figma files for hex code and keystrokes

https://www.figma.com/proto/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2592-31488&t=sdhkPVHItWMM3Xnn-0&scaling=min-zoom&page-id=187%3A12602&starting-point-node-id=2592%3A30019

bacitracin:
A11y note: Seems like the content of the cards is in flux, but when we do have the copy, let's make sure to avoid vague link titles like "Learn More" - https://www.plousia.com/blog/learn-more-or-read-more-links-bad-news-accessibility

Amberroseweeks:
I'd like to assign myself this task if no one is working on this yet. Can we consider using a <li> element and making the whole card clickable like this example? https://inclusive-components.design/cards/

bacitracin:
> I'd like to assign myself this task if no one is working on this yet. Can we consider using a < li > element and making the whole card clickable like this example? https://inclusive-components.design/cards/

Love Heydon Pickering! @marvieqa might have some opinions on the cards.

I'll assign the ticket to you, but it looks like @Moylena added the blocked tag so check to see what's up with that first.

Moylena:
@Amberroseweeks  @bacitracin  @marvieqa  I had the status as blocked because the Figma design doesn't have all the final copy!

Moylena:
@thansidwell  will drop layout updates!

thansidwell:
@Moylena @Amberroseweeks @bacitracin @marvieqa  I finished writing the content for the cards! Check out the [Figma page](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2592%3A31488&mode=design&t=8ENXrNjGR8ZzXMuG-1)

I think the links are best looking like text links with arrows suggesting you'll go to another website and it opens another tab. The number of links vary. Some have no links, most 1, some 2 and one 3. Because of that I don't think making the whole card makes sense. Plus the links allow you to inform what website you are taking them to.

[...]

PR #514, opened 2024-04-15
430-Green Content Cards on Transform Page

Amberroseweeks:
This closes issue https://github.com/CodeForPhilly/vacant-lots-proj/issues/430

I created a ContentCard component that can take in parameters like the title, body, label, description, links and an image. I added this to the Transform page and included the content from the Figma mockup. 

screencapture-localhost-3000-transform-property-2024-04-14-21_24_28


src/components/ContentCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <div className="grid gap-4 grid-cols-2 pt-5">
+            <div>
+              <span className="font-bold">{label}</span> {description}
+            </div>
+            <div>
+              <span className="font-bold">{labelUpkeep}</span> {upkeepLevel}
+            </div>
+          </div>
+          <div className="flex justify-start flex-col !items-start !text-left pt-5 !px-0 pt-5 !mx-0 !my-0">
+            {hasLink &&
```
paulhchoi, line 53: instead of having to pass in an explicit `hasLink` param, what do you think about just checking `links && links.length`?

src/components/ContentCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+              links &&
+              links.length > 0 &&
+              links
+                .slice(0, 3)
```
paulhchoi, line 57: any reason why we're slicing for the first 3 links only here?

src/images/transform/bike-rack.jpeg
paulhchoi, line 1: just a note: while I do appreciate your organization to nest all these images in a `transform` folder, the dir naming may be confusing later if we were to use these images on any other pages. No need to change now tho bc if ever that time comes, we could always just pull it out the image to the generic parent dir or something, but just a heads up

src/app/(content-pages)/transform-property/TransformPropertyPage.tsx
```diff
@@ -28,59 +51,303 @@ export default function TransformPropertyPage() {
...
+            label="Cost:"
+            description="Low"
+            labelUpkeep="Upkeep:"
+            upkeepLevel="None"
+            links={[
+              {
+                url: "https://nkcdc.org/community/cleaning-greening/community-cleanup-resources/",
+                text: "Resources on NKCDC",
+              },
+            ]}
```
paulhchoi, line 85: instead of hardcoding `upkeepLabel` and `upkeepLevel`, would it possible to match the approach you make with the `links` array, and have a param of something like:
```
<ContentCard
...
details={[
  {
    label: "Cost"
    data: "Low"
  },{
    label: "Upkeep"
    data: "None"
  }
]}

File: src/app/(content-pages)/transform-property/TransformPropertyPage.tsx [hunk 4]
```diff
@@ -28,59 +51,303 @@ export default function TransformPropertyPage() {
...
+          <ContentCard
+            hasLink={true}
+            image={
+              <Image
+                src={cleanup}
+                alt={""}
+                placeholder="blur"
+                className="rounded-t-md w-full h-44 object-cover object-center"
+              />
+            }
```
paulhchoi, line 73: for this image param, it may make it less error prone to reuse this component if we abstract away some of the details of the image component. for example, in the `<InfoGraphicSection />` component, we just require:
```
<InfoGraphicSection
...
image={{
  data: imageCleaning,
}}
..
/>
```

with some possible extra classnames passed in, and then have the `<Image />` component called within the `<InfoGraphicSection />`, where the blank alt, placeholder blurs, and default classNames are abstracted a bit

paulhchoi (changes requested):
great work @Amberroseweeks! just left a few notes, would love to see what you think

paulhchoi:
oh also, maybe @thansidwell can help with this: for tablet viewports, what do you think about having the green cards be 2 across instead of just jumping from 3 across on desktop to just 1 across on tablet + mobile? It'd help with not having the images so aggressively cropped when on tablet view

eg:
desktop view:
image

current changes tablet view:
image

thansidwell:
@Amberroseweeks It looks great! The only issue I can find it what @paulhchoi suggested. I do think it would be much better if as you sized it down it went from 3 columns to 2 to 1, rathern than 3 to 1.

Amberroseweeks:
Thank you so much for this feedback @paulhchoi and @thansidwell , i'm going to work on implementing these changes soon!

src/components/ContentCard.tsx
Amberroseweeks: I was thinking of just using 3 links as a rule but it makes much more sense to not even include this. Thank you for reviewing this component and sharing your thoughts this is all so helpful!

src/components/ContentCard.tsx
Amberroseweeks: Okay, this is now updated!

src/images/transform/bike-rack.jpeg
Amberroseweeks: Got it, no problem the transformation folder is now removed

Amberroseweeks:
@paulhchoi @thansidwell sorry I didn't tag you earlier, please review the updates when you can and thank you for your guidance!

thansidwell:
@Amberroseweeks Looks great to me! It's 3 columns to 2 to 1.

brandonfcohen1 (approved):

Issue #426, opened 2024-03-15
Content: Update Property Information table


Moylena:
On map detail page: 
- Remove "Parcel Type" from left hand side of table
- Add "Access Property" to left hand side table 

Separate table sections to be grouped as:
- Suggested Priority
- Gun Crime Rate
- Tree Canopy Gap

Then a separate table just below the above should show: 
- Access Properties
- Owner
- Zip Code
- RCO
- Council District
- Market Value
- Tax Delinquency 
- L&I violations

See Figma files for exact details: 
https://www.figma.com/proto/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2592-30687&t=sdhkPVHItWMM3Xnn-0&scaling=min-zoom&page-id=187%3A12602&starting-point-node-id=2592%3A30019

bacitracin:
A11y annotations:
![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/7329799/f9d1cb5f-794a-4306-b337-2b826010f097)

Amberroseweeks:
@bacitracin @thansidwell I am working on this now, I working on the link to the district number and RCO. I don't see an example of where the links lead to in Figma, do we have a site we are linking too specifically? Also, are the links part of the data dynamically at all? Tagging @brandonfcohen1 for thoughts too

thansidwell:
@Amberroseweeks @bacitracin @nlebovits Amber is asking where the links for District Number and RCO lead that appears in the [prototype](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2592%3A30687&mode=design&t=g2Lea4QFqnqFEToj-1). 

![CleanShot 2024-04-14 at 23 16 51@2x](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/49effbe3-0532-4af3-b62b-c946ea23ab22)

I didn't make those links, and I'm not sure who did. @nlebovits, is there value having them be links? Is there a user need there? Also where would we send them? The RCO and District website? Do they have them? Is that in the data? If this is extra database work perhaps it should a separate ticket? 

@Amberroseweeks Perhaps it's best you just skip this if we can't quickly work it out. We can make a need ticket to add them if it's needed. I don't want to stall all the work you've done.

brandonfcohen1:
[...]

Issue #505, opened 2024-04-11
Usability: Add badge to details table


thansidwell:
On the property details table, on the Suggested Priority row

- Switch out the color swatch and text
- For colored badges 

Then it will match the badge used on proporty list cards. I'm on the fence whether is should say "High Priority" or just "High."

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/8cef68c3-9690-4f38-8b6b-d55811361b7c)

Also, please remove the 1px border that appears on all badges. 

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/6318cb46-49fe-432c-b4dd-fe3379215006)

Amberroseweeks:
@thansidwell This is something I addressed in the Table Update task (https://github.com/orgs/CodeForPhilly/projects/3/views/1?pane=issue&itemId=56526628) I am just waiting to learn more about the data and what can be dynamically pulled in but this style will be applied to the table view. Here is a screenshot of the branch in progress 

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/61482332/d8396bf9-5184-4f9e-b875-b1bec3550306)

thansidwell:
@Amberroseweeks Great! Thanks.

nlebovits:
closed by #524

PR #524, opened 2024-04-17
Amberroseweeks/426-property-table-feat-505-priority-badge

Amberroseweeks:
This closes issue https://github.com/CodeForPhilly/vacant-lots-proj/issues/426 and https://github.com/CodeForPhilly/vacant-lots-proj/issues/505

Updated table content per Figma design, note that the links and tooltips within the design are not included. I also update the table style to match the figma design as well (border color and border radius)

https://www.figma.com/proto/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?page-id=187%3A12602&type=design&node-id=2592-30687&viewport=-657%2C-623%2C0.1&t=fqZvOvLyE9qv7AAV-8&scaling=min-zoom&starting-point-node-id=2592%3A30019&hide-ui=1

Screen Shot 2024-04-16 at 8 47 30 PM

**Note** I don't think we have L&I details as part of the data, I would like to check in with @brandonfcohen1 about this but I also want to get the simple updates shared. 

Right now, the L&I violations are totaled and shared as a number but in the Figma they are a list with a description and date. Is this data available and what parameters do we need to call to access it? I was not able to determine if this is in the database myself. 

figma preview on left and local branch on right

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -142,6 +142,13 @@
     @apply underline text-blue hover:text-blue-800 visited:text-purple-600;
   }
 
+  table {
+    border-collapse: collapse;
+    border-radius: 6px;
+    border-style: hidden; /* hide standard table (collapsed) border */
+    box-shadow: 0 0 0 1px #ccdce3; /* this draws the table border  */
+  }
+
   /* added webkit transform to remove icon movement in safari */
 
   .iconLink {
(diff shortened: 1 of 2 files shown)


thansidwell:
@Amberroseweeks could you also remove that 1px border on all badges? I made that little request on #505.

thansidwell (approved):
Looks great to me. 

I hoped we could remove the 1px border on badges, but I don't want to hold this up. I'll write a new ticket to do it.

Amberroseweeks:
Sorry @thansidwell I didn't notice your suggestion to remove the border on the badges, I can get to that next time! For some reason I don't get pings or notifications when i'm mentioned on GitHub

Issue #486, opened 2024-04-02
Map: Enhancement needed when pulling up properties from Search


marvieqa:
## Description

When a property is pulled up via the Search field, we don't show any indication where the property is in the map. We should point the property in the map when searched. 

- For actionable properties
- For unavailable/not actionable properties

## Additional Info

We should identify the user flow, taking in to consideration how the map works right now as a mouse user.  For example:
- When user enters an exact address of an **_available property_** and selects it:
  - Do automatically select the property on the map and display the pop-up, then show the property details on the right panel?
- When user enters an address of **_not available_** property:
  -  What does this look like?

marvieqa:
Note: We can update this ticket when we get a design.

thansidwell:
@marvieqa I would suggest this behavior. 

- When a user searches for a properity that turns out to be **vacant** -- [Figma Prototype](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=4304%3A35573&mode=design&t=Yw7lsREV4lSzMQxH-1)

  - The map shifts to the property
  - A tooltip marks the property
  - The property details are displayed on the right panel.

<img width="1057" alt="CleanShot 2024-04-02 at 23 58 40@2x" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/bd59bf8d-b1cb-4ed0-80d3-b34c2c0df8d6">

- When a user searches for a properity that turns out to be **not vacant** -- [Figma Prototype](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=4304%3A35330&mode=design&t=Yw7lsREV4lSzMQxH-1)
  - The map shifts to the property
  - A tooltip marks the property
  - The right panel doesn't change, but stays the list
    
<img width="1058" alt="CleanShot 2024-04-02 at 23 58 33@2x" src="https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/0a441cd0-a047-426a-8914-73f7d43ba44a">

jgrimes86:
Hello @Moylena, I think I can address this one.

nlebovits:
closed by #532

PR #532, opened 2024-04-18
486: Map behavior after property search

jgrimes86:
This addresses issue #486 

When a user searches for a property in the map's search bar, the map shifts to the property and a tooltip appears marking the property.

If the searched-for property is vacant, the side panel changes to the property details.  If the property is not vacant, the side panel remains in the list view or (if it was previously showing another property's details) returns to the list view.

@thansidwell - I don't see any details about whether a zoom level should be set for the search result, so as of now the zoom level does not change when a property is searched.

Also, I thought I saw an assigned ticket that was handling the styling of the map tooltip.  I can't find it now and don't see changes on staging site, so if I was imagining that please let me know and I'll work on the tooltip styling.

@paulhchoi and @brandonfcohen1 - For properties that are out of view when they are searched, I had some trouble checking whether they are in the dataset of vacant properties .  My solution was to delay calling the 'handleMapClick' function (which I'm re-purposing for the search) with a setTimeout() with a 0.5 second delay.  I don't know if that's best practice, or enough time/too little time, but I couldn't get it to work otherwise.

### DIFF
--- a/src/app/find-properties/[[...opa_id]]/page.tsx
+++ b/src/app/find-properties/[[...opa_id]]/page.tsx
@@ -214,6 +214,7 @@ const MapPage = ({ params }: MapPageProps) => {
               </div>
               {isLinkedPropertyParsed ? (
                 <PropertyMap
+                  featuresInView={featuresInView}
                   setFeaturesInView={setFeaturesInView}
                   setLoading={setLoading}
                   selectedProperty={selectedProperty}
--- a/src/components/PropertyMap.tsx
+++ b/src/components/PropertyMap.tsx
@@ -49,6 +49,11 @@ import { Info } from "@phosphor-icons/react";
 import { centroid } from "@turf/centroid";
 import { Position } from "geojson";
 
+type SearchedProperty = {
+  coordinates: [number, number],
+  address: string,
+};
+
 const MIN_MAP_ZOOM = 10;
 const MAX_MAP_ZOOM = 20;
 const MAX_TILE_ZOOM = 16;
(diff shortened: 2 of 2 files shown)


brandonfcohen1 (changes requested):
hey @jgrimes86 - check out how I do it with the routing here: https://github.com/CodeForPhilly/vacant-lots-proj/blob/129446d8660c0a149b0ab244561ce4054aa9942b/src/app/find-properties/%5B%5B...opa_id%5D%5D/page.tsx#L58-L117

jgrimes86:
Thank you for the pointer @brandonfcohen1.  I think I have it working with now with a useEffect.  I'm going to test it out a little more tomorrow morning when I'm fresh, then I should have an improved version for you to review.

thansidwell:
@jgrimes86 This works great!

And yes, the zoom should just stay as it is. That's the same behavior we use as someone clicks on properties on the map or the list. We might consider it after testing it out a lot, but this seems best for now.

jgrimes86:
@brandonfcohen1 - I've changed the function for handling map searches to a useEffect and added state for `searchedProperty` to store the searched property info.  The useEffect runs after updates to featuresInView and selectedProperty, and it returns nothing if featuresInView doesn't exist, there is already a selectedProperty, or searchedProperty's address is '' (which I'm using as the default/non-searched property value).

Since a search's selectedProperty is being set by the new useEffect, I've changed `handleMapClick` back to its previous form.

brandonfcohen1 (approved):
looks great- will let @thansidwell make a new ticket if he thinks the zoom level should be changed

Issue #502, opened 2024-04-10
Styling: Improve property tooltip


thansidwell:
Right now the current property tooltip on the map

- Blends into the map since it's white
- Doesn't match the style of the rest of the site
- The close icon is awardly placed

Change the tooltip to this styling. 

- Dark background, white text, rounded edges, new close icon, close icon far right inline with the text

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/7c752de8-e374-4e15-a729-f34c6a3a2917)

It should use the [tooltip component](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=4304%3A35855&mode=design&t=EhbuNgZAuhBTthHD-1) or be styled to look like it. I have it built out with different variations. With and without a tick/tail/triangle and with and without a close icon.

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/79cf2a02-8e3d-4f43-849a-595fd0278e8f)

Here's the [prototype page](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=4304%3A35573&mode=design&t=EhbuNgZAuhBTthHD-1) where the property tooltip appears.

ajtw7:
@nlebovits assign to me.

Moylena:
@ajtw7  assigned!

nlebovits:
closed by #555, #562

PR #542, opened 2024-04-18
Styling: Improve Property Tooltip (502)

ajtw7:
@brandonfcohen1 @paulhchoi 

This PR is a fix for issue #502 

The tooltip and pointer are being affected by this global styling import in `global.css`

`import '~mapbox-gl/dist/mapbox-gl.css';`

I added brute force CSS rules for the affected classes to the bottom of the `global.css`. 
Not entirely sure if it's the best—or most efficient—solution.

I'm more than happy to hear other recommendations.

Also, I added a bit of styling inside of the`Popup` component, specifically for the 'X'.
Please ignore the unused `PropertyMap.module.css` import.


src/app/components/PropertyMap.tsx
```diff
@@ -23,41 +23,41 @@ import Map, {
...
+      'Medium',
+      '#FFD700', // Gold
+      'Low',
+      '#B0E57C', // Light Green
       //"Medium Priority",
       //"#FF8C00", // Dark Orange
       //"High Priority",
       //"#B22222", // FireBrick
       //"Top 1%",
       //"#8B0000", // Dark Rednp
```
paulhchoi, line 54: could we remove this if not being used?

src/app/components/PropertyMap.tsx
```diff
@@ -328,17 +328,41 @@ const PropertyMap: FC<PropertyMapProps> = ({
...
-            <div>
-              <p className="font-semibold text-md p-1">
-                {popupInfo.feature.address}
-              </p>
+            <div
+              style={{
+                display: 'flex',
+                justifyContent: 'space-between',
+                flexWrap: 'nowrap',
+              }}
```
paulhchoi, line 336: just a nitpick, but updating this to tailwind to keep it consistent with the codebase would be nice

src/app/components/PropertyMap.tsx
```diff
@@ -328,17 +328,41 @@ const PropertyMap: FC<PropertyMapProps> = ({
...
+                  strokeWidth={1.5}
+                  stroke="currentColor"
+                  className="w-4 h-4"
+                >
+                  <path
+                    strokeLinecap="round"
+                    strokeLinejoin="round"
+                    d="M6 18 18 6M6 6l12 12"
+                  />
+                </svg>
```
paulhchoi, line 357: any reason you didn't use a Phosphor Icon as we used across the app here? if not tho, I'd prefer to take raw svg's and label them as a component so it's more easily parsable in the future (eg: adding `const XIcon = () => <svg>...` and then using `<XIcon />` here)

paulhchoi (changes requested):
added small nits, but good work overall! I'm fine with the brute force css, seems like the main way we've been customizing the map lol.  Other than that, the prototype has "Title Case" for the addresses, a bit thinner font, smaller padding around the left/right/top, and corners that are fully rounded — could you take another pass at adding these styles to the tooltips?

image

vs

image

ajtw7:
@paulhchoi got it 👍🏾. Thanks for the feedback. I'll get on the changes.

thansidwell:
Great work, @ajtw7! Thanks for pointing out those issues, @paulhchoi. 

I'll illustrate what the padding should be, since it might be hard to get in Figma. 

CleanShot 2024-04-19 at 12 33 46@2x

The text uses the body-sm text style which is 14px, normal weight, and title case.

Moylena:
@ajtw7 Hi - looks like a conflict must be resolved? Can you hop on this today?

src/app/components/PropertyMap.tsx
ajtw7: Done.

src/app/components/PropertyMap.tsx
ajtw7: All good. Done.

src/app/components/PropertyMap.tsx
ajtw7: No reason in particular. I wasn't sure of the global library we were using and `Favicon` is always good boilerplate. I'm using an icon from `Phosphur` now.

ajtw7:
@paulhchoi I moved the `toTitleCase` function into a `utilities` folder; thought it might be more efficient since `PropertyMap.jsx` will need it as well. Either way, let me know if I need to make any additional changes.

@thansidwell thank you for the specifics on padding!! I just saw your message. I think I got it pretty close. I'll wait for the final word.

thansidwell:
@ajtw7 @paulhchoi  The Preview shows a version of the website that's pretty out of date. I'm not a developer but I would guess that's why there's conflicts. Can you update to the latest?

ajtw7:
@thansidwell I think you're right. Looking into it. So far, haven't been able to pull/fetch latest version.

ajtw7 (commented):
Same file. New location.

PR #545, opened 2024-04-19
config and doc changes for a more robust backend setup

zigouras:
I got the frontend and backend working on my WSL Ubuntu setup on Windows.  I found some small areas for improvement in configuration and documentation as I worked through some setup issues with Brandon.


brandonfcohen1 (changes requested):
this is awesome, thank you so much @zigouras 

question- did you have any trouble with the CORS setup on the bucket? should we add some documentation about that?

Moylena:
@zigouras  can you make the change requested? Then I think this will be all set

brandonfcohen1 (approved):

zigouras:
> this is awesome, thank you so much @zigouras
> 
> question- did you have any trouble with the CORS setup on the bucket? should we add some documentation about that?

It wasn't CORS after all, it was simply the permissions on my bucket in GCP.  I had to let the 'allUsers' principal have read access to the bucket.  This makes it publicly available on the web the same as the prod bucket.

PR #559, opened 2024-04-21
RCO Filter working and general cleanup

brandonfcohen1:
I got the RCO filter working. In the filter config, you can set a boolean `useIndexOfFilter` = `true` to allow for the use of this `indexOf` style filter.

https://www.loom.com/share/27da7187b3004fffac5e3c0714224092

I also noticed that RCO was showing neighborhood in the property detail. I fixed that and added the RCO list:

image

I also cleaned up some formatting


thansidwell (approved):
It works beautifully!

PR #570, opened 2024-04-21
Filter fix 42124

brandonfcohen1:
fix issue with RCO filters when others are activated

Screenshot 2024-04-21 at 6 00 06 PM


thansidwell (approved):
I put it through its paces and I can't find any issues. Great job, @brandonfcohen1!

Issue #582, opened 2024-04-22
Search doesn't work


thansidwell:
You can't click into the search field to use it.

https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/d1a4fae1-4fc9-45ca-a38d-c05f170d9f64

jgrimes86:
Adding a `pointer-events` attribute to the following section of globals.css (starting on line 225) seems to fix the issue.
```
#main .maplibregl-ctrl-top-right {
  top: 20px;
  right: 20px;
  pointer-events: bounding-box;
}
```
@thansidwell is it OK if I make a pull request so you can check out this change?

thansidwell:
Definately!

thansidwell:
Just a thought. It seems strange a bit strange to add something to make it work even though it worked before. It seems like we should be taking something away whatever is obstructing it.

But I'm good whatever works!

jgrimes86:
@thansidwell I think the issue is that `@import "~mapbox-gl/dist/mapbox-gl.css";` was removed from globals.css.  That css file includes a 'pointer-events: auto' attribute for all map controls, along with styling for other elements.  I assume it was removed to allow custom styling of other map features, so I'm afraid of adding it back in and breaking other things.

thansidwell:
Ah, that makes sense now!

thansidwell:
@jgrimes86 Also the map legend tooltip doesn't work either. I assume for a similar reason. 

https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/5c9608cf-2c59-4b57-8252-3235f91fdefa

jgrimes86:
Yes, that was the same issue.  I didn't know that the tooltip had been interactive.  I'm going to push a commit to my PR request branch to address that issue.

Also, I made one more small change to fix a bug with the map click function. Clicking the map had been re-directing to the last searched property.

nlebovits:
closed by #586

PR #586, opened 2024-04-22
Bug Fix: Map Search Field Not Clickable

jgrimes86:
This PR addresses two bugs where you could not click into the map's search field and you could not click the map tooltip. It fixes the bugs by adding `pointer-events: bounding-box' to globals.css for the classes for each of those features:
- maplibregl-ctrl-top-right
- maplibregl-ctrl-bottom-left

I've also added `setSearchedProperty({...searchedProperty, address: "" })` to the map `handleClick` function.  Without that, if the map is clicked on a non-vacant property, the map changes the popup tooltip to the last searched property.

This closes issue #582

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -225,6 +225,7 @@
 #main .maplibregl-ctrl-top-right {
   top: 20px;
   right: 20px;
+  pointer-events: bounding-box;
 }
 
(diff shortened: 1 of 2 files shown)


nlebovits:
Thanks @jgrimes86! @all-contributors, please add @jgrimes86 for code

PR #595, opened 2024-04-24
561/ Fix Page Content Spacing

AZBL:
This changes makes the ticket's requested styling changes.

For the About page:
- Increases subheader text size to body-lg
- Decreases margin between image and subheader to 48px

For the Transform Property page:

- Increases margin between subheader and image to 48 px.
- does NOT change anything with margin above page header as that was done seperately

### DIFF
--- a/src/app/(content-pages)/about/AboutPage.tsx
+++ b/src/app/(content-pages)/about/AboutPage.tsx
@@ -10,7 +10,7 @@ export default function AboutPage() {
   return (
     <>
       <h1 className="heading-3xl font-bold mb-6">About This Project</h1>
-      <p className="body-lg mb-12">
+      <p className="body-lg mb-6">
         Philadelphia has a gun violence problem. Clean & Green Philly empowers
         Philadelphians to take action to solve it.
       </p>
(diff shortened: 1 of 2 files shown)


AZBL:
@thansidwell @Moylena @paulhchoi

thansidwell:
The About page had a larger gap than it should. About 74px instead of 48px. I think it's because of a break tag. 

CleanShot 2024-04-24 at 10 03 44@2x

AZBL:
@thansidwell @paulhchoi 

It should be fixed now. Seems like the <br /> has a line height of 24 px so i just reduced the bottom margin of the subheader <p/> to 24 px. Hopefully that solves it.

thansidwell:
@AZBL The preview still shows the problem for me. 

Does that `<br>` tag need to be there? Can we remove it? That fixes the problem for me. 

https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/9f6ebabb-3aa4-407d-bc17-cc2ddb178e5e

AZBL:
Hey @thansidwell it looks like the 2nd video you posted doesn't include the changes from my most recent commit. In the video, for the subheader p-tag it still says mb-12 (which is 48 px of spacing which then gets added onto the 24 px from the break) tag. The more recent commit I had changed it to mb-6 which reduces it to 24 px of spacing which gets added onto the 24 from the break. It looks fine in my browser.

I'm not 100% sure if the break tag is necessary but I didn't want to just delete it.

Let me know if you're still not able to see the changes I made and I'll just push it again.

thansidwell (approved):
I'll just approve this. I can't see it in the preview, but I trust you it's fixed with it being such a small issue. Thanks!

Issue #581, opened 2024-04-22
Simplify & combine "Get Access" filter


thansidwell:
The filter has some usability issues.

- These different filters are actually different methods of getting access, but they are grouped in seperated field groups likely to created confusion. 
  - Buy Property, Land Bank, Private Land Use Agreement, Conservatorship Eligible, Side Yard Eligable and Tactical Urbanism
- Most methods like "Land Bank, Private Land Use Agreement, and Conservatorship are likely hard to understand for the lay person. They need more explaination for people to understand if they are good options for them. 
- The logic of how these filters work isn't the same which will likely also confuse people. 
  - Filtering by Buy Property Land Bank, Private Land Use Agreement will show operties where the option is the **_best option_**.
  - Filtering by Conservatorhips, Side Yard, and Tactical Urbanism will show properties where the option is **_avaliable_**. 

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/e99e6ed8-1ac1-43dd-9c2f-0cda40dd06cd)

Make these these changes to address those issues

- Create a new selectable panel for each filter 
  - Add an icon for each to help with recognition
  - Add heading and description to bring clarity to what they mean
  - Clicking on it changes it to a selected state with a blue background a check mark
  - Build in the Hovering and Pressing states as well [View prototype](https://www.figma.com/file/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=6340%3A41341&mode=design&t=Pbc7oHPHsvr5J4Sr-1)

- Filtering by these options would filter to properties where those options are the either "Avaliable" or the "Best Option." 

- Add a "Find only when it's best option" badge button
  - Clicking this button would filter to properties where those options are the "best option"

- Delete the old sections

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/f2af258f-9112-46d7-b2b7-214d2f62c3ec)

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/13c6c0a3-16d3-4142-8424-95d159feae24)

All these changes harmonizes with the improved "Get Access" section, making is 

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/6d4a7353-9047-4900-a5f7-bee58885a41e)

jgrimes86:
@thansidwell I'd like to work on this issue, but I have a question about the 'Find only when its the best option' button.

[...]

PR #600, opened 2024-04-27
581 simplify get access filter

jgrimes86:
This PR request combines the "Access Process", "Conservatorship Eligible", "Side Yard Eligible", and "Tactical Urbanism" into one, new filter section titled "Get Access" which contains six selectable panels: three for access process (private land use agreement, buy from owner, and get from land bank) and one each for conservatorship, side yard, and tactical urbanism.

This PR addresses issue #581 

@thansidwell and @paulhchoi, I have made frontend changes for review in this PR

Some things to note for the reviewers' consideration:

- The 'Find only when its the best option' button from the prototype has not been included because it doesn't seem to add functionality to the filter.  See discussion in issue #581 

- The new panel filter has different functionality than the previous Yes/No button filters for conservatorships, side yards, and tactical urbanism.  In the panel filter, selecting one of those options filters for the "Yes" condition of those options.  Deselecting the option returns the map to showing properties that satisfy either the Yes or No conditions for that option.  A user cannot choose to see the properties where those options aren't available.

- The "Get Access" filter reuses the icons and titles from the "Getting Access" section of the property detail view.  The titles don't exactly match the "Get Access" panel titles in the prototype, but I thought reusing them would be good for continuity.  I'm no longer sure that level of continuity makes a difference, so I won't argue against changing the titles.  I kept the matching titles for this PR for two reasons: it gives me more mileage out of extending `PanelFilterOptions`, which is a cool new trick for me, and I thought it would be easier to point out the difference in phrasing this way.  I'll change the panel titles to what was planned in the prototype if that is preferred.

- There is not a smooth transition in color between the default, hover, and selected states of the panels, especially when a selected panel is deselected.  I think I see the same issue happening on the other filter buttons, but it is more pronounced on the large panels.  Any suggestions on how to reduce or eliminate the flickering effect would be appreciated.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -297,7 +297,7 @@ a .bg-color-none {
 }
 
 .panelDefault:hover {
-  @apply ease-in bg-gray-200;
+  @apply ease-out bg-gray-200;
 }
 
 .panelDefault:active {
@@ -309,7 +309,7 @@ a .bg-color-none {
 }
 
 .panelSelected:hover {
-  @apply ease-in bg-blue-300;
+  @apply ease-out bg-blue-300;
 }
 
 .panelSelected:active {
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -292,6 +292,30 @@ a .bg-color-none {
   @apply bg-blue-300;
 }
 
+.panelDefault {
+  @apply flex flex-row items-center rounded-md p-3 space-x-3 bg-gray-100 text-gray-900;
+}
+
+.panelDefault:hover {
+  @apply ease-in bg-gray-200;
+}
+
+.panelDefault:active {
+  @apply ease-in bg-gray-300;
+}
+
+.panelSelected {
+  @apply flex flex-row items-center rounded-md p-3 space-x-3 bg-blue-200 text-blue-800;
+}
+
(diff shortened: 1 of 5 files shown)


thansidwell:
Hey @jgrimes86. Great work! The only issue I could find was 

- Adding a 12px gap between the icon and text

https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/5bf5bf6a-7aba-466b-be6e-59bc0ef8ceb4

I'll write another ticket to fix the issue about the flashing background, the fluctuating numbers, and making the language even more consistent.

jgrimes86:
@thansidwell I adjusted the space between the icon and text.

Thanks for making a separate ticket for the strobing color and numbers.  I didn't know how to proceed with the colors, and the numbers issue is new to me.  I don't see the fluctuating numbers in my local version or the vercel preview, so I don't know what would have caused that.

thansidwell (approved):
@jgrimes86 I can’t see the change in the preview. You have to redeploy it or something. But I trust you did, so I’ll save you the time and just approve this. Thanks for all your work on this! 

@brandonfcohen1 Please merge this to staging and live whenever you can! I think it makes really improves the usability significantly.

PR #587, opened 2024-04-23
415: Feature: Save properties

CodeWritingCow:
## Description
This PR adds a "Save" button to the property details UI. When users click the button, it saves the property to their Web browser's local storage. They can "unsave" the property by clicking the button again.

This PR also adds a "Saved" button to the property list UI. When users click the button, it filters the list of properties to those saved in local storage.

## Screenshots

Screenshots coming soon.

Closes #415

PR #607, opened 2024-04-28
415: Fix Save button issues

### LINKED ISSUE #415: Feature: Save properties
Currently there is no way to save a property in the property detail page

Please add the icon in the top right hand of the property detail card, on the same line as the "back" link. Include the default and active state as a part of this ticket 

Image

Image

When user clicks this new icon, all properties with this "save" flag will appear. It's another filtering mechanism. 

--- 

See Figma file for details: 
https://www.figma.com/proto/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?type=design&node-id=2592-30687&t=sdhkPVHItWMM3Xnn-0&scaling=min-zoom&page-id=187%3A12602&starting-point-node-id=2592%3A30019

CodeWritingCow:
## Description

This PR fixes the following issues from PR #587 with the feature, "Save properties to local storage":

- UI displays wrong count for "# Properties in View." This PR fixes the issue by showing `savedPropertyCount` instead of  `featureCount` if a user is viewing a list of saved properties.
- Filter button is changing to a selected state and showing a filterCount of "(1)" if a user is filtering by saved properties. This PR fixes the issue by overriding this default behavior and excluding OPA_ID from `filterCount`. Filter button now remains unchanged if user views a list of saved properties. 

## Screen recording

https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/9e105ee9-6266-4ef3-8640-71d17b0752a1

Closes #415

### DIFF
--- a/src/app/find-properties/[[...opa_id]]/page.tsx
+++ b/src/app/find-properties/[[...opa_id]]/page.tsx
@@ -298,6 +298,9 @@ const MapPage = ({ params }: MapPageProps) => {
                   setSelectedProperty={setSelectedProperty}
                   setIsStreetViewModalOpen={setIsStreetViewModalOpen}
                   shouldFilterSavedProperties={shouldFilterSavedProperties}
+                  setShouldFilterSavedProperties={
+                    setShouldFilterSavedProperties
+                  }
                   smallScreenMode={smallScreenMode}
                   updateCurrentView={updateCurrentView}
                 />
--- a/src/components/PropertyDetailSection.tsx
+++ b/src/components/PropertyDetailSection.tsx
@@ -52,6 +52,7 @@ interface PropertyDetailSectionProps {
   setSelectedProperty: (property: MapGeoJSONFeature | null) => void;
   setIsStreetViewModalOpen: Dispatch<SetStateAction<boolean>>;
   shouldFilterSavedProperties: boolean;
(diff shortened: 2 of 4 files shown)


CodeWritingCow:
When a user switches view from saved properties back to all 30,000+ properties, sometimes an incorrect count for "# Properties in View" appears briefly and disappears before the correct count. This happens while the map is reloading, and `featureCount` is being recalculated. I haven't figured out how to eliminate this behavior completely yet. Just want to call that out. Nonetheless, the UI does display the correct "# Properties in View" after the map finishes loading.

thansidwell:
@CodeWritingCow Can I ask for just one more thing I realized should happen? Very small. 

- Move Save first and Share second. That matches the order of Save on the list and Save will probably be used more than Share. 

CleanShot 2024-04-28 at 19 02 34@2x

CodeWritingCow:
> @CodeWritingCow Can I ask for just one more thing I realized should happen? Very small.
> 
> * Move Save first and Share second. That matches the order of Save on the list and Save will probably be used more than Share.
> 
@thansidwell No problem! I just updated this PR so Save button precedes Share button. See screenshot:

Screen Shot 2024-04-28 at 7 20 24 PM

thansidwell (approved):
Thanks for all the fixes!

Issue #603, opened 2024-04-28
A11y: On narrower screens the Filter button has no text & announces as just "button"


bacitracin:
## Describe the task
When using a screenreader to navigate through the property page buttons they should always announce what they are, "button, Filter" or "button, Download". On desktop screen widths we show the text within the button, and that text is announced. When the screen is narrower, the text disappears. The download button has an aria-label, so it announces. The Filter button needs an `aria-label="Filter"` as well.

## Acceptance Criteria

- [ ] Go to the Find Properties page
- [ ] Resize the screen so that it text in the button disappears.
- [ ] Turn on screenreader. Navigate to the Filter button.
- [ ] It announces as "Filter button"

10234567Z:
I am ready to take this issue, I can tackle this

bacitracin:
@10234567Z Assigned!

PR #608, opened 2024-04-29
Fix a11y issue with Filter button (#603)

10234567Z:
+ Fix the issue of the Filter button not being read by screen readers on small screens


bacitracin (approved):
👍

nlebovits:
closes #608

Issue #594, opened 2024-04-24
Fix property list controls


thansidwell:
Fix the styling of the property list controls

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/8fb5d4a3-ac06-448c-9573-1a642cc8de8f)

To match the prototype

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/5247928e-a228-4e01-a774-557a808617ff)

And match the single property controls

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/941fe1b9-364e-4754-a90e-87e2c09d76e1)

 - Ensure the padding matches the prototype
 - Remove the bottom border line
 - Remove the extra padding around the text

![Image](https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/7e0f7df6-ab7f-468c-a503-2094887a8792)

Moylena:
Is there an associated PR for this? I don't see anything in this audit trail @AZBL

AZBL:
> Is there an associated PR for this? I don't see anything in this audit trail @AZBL

There should be I though lt it was done but I’ll check later

AZBL:
> Is there an associated PR for this? I don't see anything in this audit trail @AZBL

@Moylena 
#611 
https://github.com/CodeForPhilly/clean-and-green-philly/pull/611

Moylena:
Thanks, please tag the ticket you are closing on each PR so it tracks back!

PR #611, opened 2024-04-29
Fix Property List Controls - 594

AZBL:
This change fixes the styling of the property list controls on the Find Properties page. 

- It should now match prototype.
- It removes the bottom border line.
- Removes left padding from [number] properties in view from large screens so there should only be 24 px of padding now
- Keeps left padding for [number] properties in view for small and medium screens so there is an extra 16px of space (otherwise when it switches to 1 property property img per row and the download/filter buttons change to icons only the alignment looks off). I included how this looks with the extra padding and without the extra padding. 

With Extra Padding:
with_padding

Without Extra Padding:
without_padding

- Removes right padding from the Download button so there should only be 24 px of padding now


thansidwell:
Hey @AZBL. Just one little issue and then it's all good. The hover background of the property cards' top padding gets cut off. Dropping it down like 8px or so would fix it. 

CleanShot 2024-04-30 at 00 09 35@2x

AZBL:
@thansidwell 
I added 8 px of top margin to the div containing the Table so now the hover background does not get cut off.

thansidwell (approved):
Approved. @AZBL said he fixed the small issue I found.

Moylena:
#594

Issue #601, opened 2024-04-27
Strobing buttons and numbers in filter


thansidwell:
In the filter when you click on one of the button badges the background flashed or strobes light, dark, light, dark, light, dark. For most people this just feels janky. For people with epilepsy strobing effects can induce seizures, so I would assume in this small amount it would likely be at least unsettling for them. 

Also as you filter the number of properity jumps to one number and switches to another. 

These two issues might be related or might not be. 

https://github.com/CodeForPhilly/vacant-lots-proj/assets/1965986/ea22ea01-543c-490f-a7da-efc7929f248d

jgrimes86:
@thansidwell the strobing button color has been on my mind, so I've been looking into a fix.  I think I've resolved that issue.  I don't know if this also affects the numbers issue, because I didn't see that on my local version of the app.

Can I make a pull request so you can see how it works on your end?

https://github.com/CodeForPhilly/clean-and-green-philly/assets/140654473/59e1062d-e0f5-4d50-aab7-a8ef022f9305

PR #617, opened 2024-04-30
Bug Fix: Correct flashing of filter button color on select & hover

jgrimes86:
This addresses the strobing color issue in #601 

The issue was caused by the filters being re-rendered twice whenever they were selected or deselected.  This PR resolves the issue by putting the `filter` array inside a useMemo that only updates when the `selectedKeys` or `selectedPanelKeys` states change.

https://github.com/CodeForPhilly/clean-and-green-philly/assets/140654473/c583d2c7-1efb-408f-b2d1-dcaf995b69b0

### DIFF
--- a/src/components/Filters/DimensionFilter.tsx
+++ b/src/components/Filters/DimensionFilter.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { useState, FC } from "react";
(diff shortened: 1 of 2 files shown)


thansidwell (approved):
Nice work figuring this out!

PR #636, opened 2024-05-16
Fix #602: Tooltip for share button when tabbing (a11y) 

### LINKED ISSUE #602: A11y: Tooltip does not show for Share button on keyboard focus/activation
## Describe the task
When you hover over the "Share" button with the mouse, a tooltip appears that tells the user they can copy the link. After clicking on the button, a tooltip tells the user that the action was successful. These tooltips should also be visible on keyboard focus.

## Acceptance Criteria

- [ ] Open a property page
- [ ] TAB over to the "Share" button.
- [ ] On keyboard focus, the "Copy link" tooltip should be visible.
- [ ] Press ENTER to activate the button.
- [ ] The "Link copied" tooltip should be visible.

## Additional context

* Possibly helpful: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible and https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within

amyyeung17:
## Description
This PR addresses issue #602.

## Testing/Proof

https://github.com/CodeForPhilly/clean-and-green-philly/assets/28910543/36f37ab5-d6dc-40be-9ee2-b9dfe65282f4

### DIFF
--- a/src/components/SinglePropertyDetail.tsx
+++ b/src/components/SinglePropertyDetail.tsx
@@ -208,7 +208,6 @@ const SinglePropertyDetail = ({
             }}
           >
             <ThemeButton
-              tabIndex={0}
               color="tertiary"
               label="Share"
               startContent={<Share />}
--- a/src/components/SinglePropertyDetail.tsx
+++ b/src/components/SinglePropertyDetail.tsx
@@ -208,6 +208,7 @@ const SinglePropertyDetail = ({
             }}
           >
             <ThemeButton
+              tabIndex={0}
               color="tertiary"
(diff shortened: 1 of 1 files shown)


amyyeung17:
@bacitracin I realized there was an edge case that I didn't cover. Do you want the tooltip to stay visible even if the mouse moves in and out of the button area, as long as the button is in focus?

bacitracin:
@amyyeung17 No, I think it's unnecessary to also try to cover that edge case. Good catch, tho!

bacitracin (approved):

PR #644, opened 2024-05-21
604 /  Apply Hover Styling on Property Cards when Keyboard Focus is on Cards

AZBL:
Now, when a user is on the Find Properties page and uses their keyboard to tab to focus on a specific card, the background of the entire card will change to gray similar to hovering over the card with a mouse. Previously the background remained white. 

With change:
Screenshot 2024-05-21 at 2 33 24 PM


AZBL:
@bacitracin

Issue #629, opened 2024-05-09
Limit Search to Philadelphia


thansidwell:
When users serach there can be many results that are not located in Philadelphia. 

![Image](https://github.com/CodeForPhilly/clean-and-green-philly/assets/1965986/31060b8e-c124-44c1-a74f-d82fb43385f7)

Please restrict the search to only addresses found in Philadelphia.

jgrimes86:
@Moylena or @thansidwell can this be assigned to me?

CodeWritingCow:
@jgrimes86 I just assigned the ticket to you. Thanks for tackling this!

PR #683, opened 2024-06-07
629: Limit Search to Philadelphia

jgrimes86:
This PR adds a bounding box and filter to the mapbox geocoder to limit search results to addresses located in Philadelphia.

The filter limits search results to items that have a place feature of 'Philadelphia' or a district feature of 'Philadelphia County'.  The bounding box limits search results to a rectangular area around Philadelphia, PA so that the search doesn't return results from other places named Philadelphia, such as Philadelphia, Mississippi.

This closes issue #629 

image


nlebovits:
Thanks @jgrimes86! @CodeWritingCow PR review?

Issue #685, opened 2024-06-08
Bug: Selected Filter Panels Lose Their 'Selected' Appearance


jgrimes86:
**Describe the bug**
When a panel in the 'Get Access' section is selected, the color changes to its selected state.  However, if you toggle to the property card view and then back to the filter panel view, the panel no longer appears to be selected, even though the filter is still active.

**To Reproduce**
1. Go to the filter view.
2. Click on one or more panels in the 'Get Access' section to active those filters.
3. Exit the filter view so that you can see the property cards, then go back to the filter view.
4. See that the selected panels appear to not be active

**Expected behavior**
Filter panels will remain in their 'selected' state when toggling between the property card view and filter view

**Screenshots**
This is what a selected panel looks like when first selected.  It should continue to look like this as long as the filter is active:
![Filter showing panel selected](https://github.com/CodeForPhilly/clean-and-green-philly/assets/140654473/22d3313b-3c04-4ce6-b6c9-8acf9ca28bfa)

This is what the the selected panel currently looks like after leaving and returning to the filter view:
![Filter not showing panel selected](https://github.com/CodeForPhilly/clean-and-green-philly/assets/140654473/790e582c-5e0b-4649-936e-ade2893d5ac2)

jgrimes86:
Also, I think I know how to correct this, so I would like to be assigned to it.  Thank you.

nlebovits:
Assigned @jgrimes86!

PR #688, opened 2024-06-08
685: Bug Fix: Selected Panels 'Selected' State

jgrimes86:
This addresses issue #685 

The selectedPanelKeys state, which controlled the appearance of selected panel filters is not saved when you toggle back and forth between the filter view and property cards view.  To keep the panel filters correctly selected after toggling the view, the initial 'selectedPanelKeys' state is now generated from appFilter.

This is an example of the corrected panel filter appearance after toggling between the filter view and property cards view:
image

### DIFF
--- a/src/components/SidePanelControlBar.tsx
+++ b/src/components/SidePanelControlBar.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { FC, useRef } from "react";
+import React, { FC, useMemo, useRef } from "react";
 import { BarClickOptions } from "@/app/find-properties/[[...opa_id]]/page";
 import { BookmarkSimple, DownloadSimple, Funnel } from "@phosphor-icons/react";
 import { ThemeButton } from "./ThemeButton";
@@ -32,20 +32,20 @@ const SearchBarComponent: FC<SidePanelControlBarProps> = ({
   const savedRef = useRef<HTMLButtonElement | null>(null);
   const { dispatch, appFilter } = useFilter();
 
-  const filterCount = () => {
+  const filterCount: number = useMemo(() => {
     let count = 0
     for (let property of Object.keys(appFilter)) {
-      if (property === "access_process") {
-        count = count + appFilter[property].values.length
-      } else {
-        count++
+        if (property === "access_process") {
+          count += appFilter[property].values.length
+        } else {
+          count++
+        }
       }
-    }
-    if (shouldFilterSavedProperties) {
-      count--
-    }
+      if (shouldFilterSavedProperties) {
+        count--
+      }
     return count
-  }
+  }, [appFilter])
 
   const onClickSavedButton = () => {
     let propertyIds = getPropertyIdsFromLocalStorage();
@@ -115,7 +115,7 @@ const SearchBarComponent: FC<SidePanelControlBarProps> = ({
             label={
               <div className="lg:space-x-1 body-md">
                 <span className="max-lg:hidden">Filter</span>
-                {filterCount() !== 0 && <span>({filterCount()})</span>}
+                {filterCount !== 0 && <span>({filterCount})</span>}
               </div>
             }
             onPress={() => {
@@ -125,7 +125,7 @@ const SearchBarComponent: FC<SidePanelControlBarProps> = ({
 
               updateCurrentView("filter");
             }}
-            isSelected={currentView === "filter" || filterCount() !== 0}
+            isSelected={currentView === "filter" || filterCount !== 0}
             startContent={<Funnel />}
             className="max-lg:min-w-[4rem]"
             data-hover={false}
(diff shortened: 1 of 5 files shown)


jgrimes86:
@CodeWritingCow I think this PR is one for you to review

CodeWritingCow:
Hi @jgrimes86 I tested your PR in my local env. When I toggled between the filter and property card views, the user's filter choices do persist. However, I encountered two unexpected behaviors:

https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/baca9756-2d0d-450a-aa2a-117ad60db832
1) Filter count mismatch: For example, if I select all 6 filters in the "Get Acess" section, the UI displays the count as 4.

https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/65d915ee-492c-4139-8c14-3d561c6e9eba

2) If I navigate from "Find Properties" page to another page, then back to "Find Properties," then the user's filter choices don't persist. (In other words, the user's selected filters become "unselected.")

CodeWritingCow:
@jgrimes86 can you take a look at the first behavior?

As for the second behavior, it might be caused by an issue unrelated to your current ticket. What do you think?

CodeWritingCow (commented):
@jgrimes86 can you take a look at the first behavior?

As for the second behavior, it might be caused by an issue unrelated to your current ticket. What do you think?

jgrimes86:
> @jgrimes86 can you take a look at the first behavior?
> 
> As for the second behavior, it might be caused by an issue unrelated to your current ticket. What do you think?

@CodeWritingCow I've pushed two commits to address both those issues.

For the first behavior, I needed to update filterCount to reflect that three of the 'Get Access' buttons are different values of one 'access_process' filter.

I think the second behaver is related to this ticket in that it is annoying that selected filters don't persist when you're using the website.  But it might also be considered unrelated because it is caused by how filter context is saved, not by how the filter context is handled within the side panel components.  Fixing that behavior seems to be a pretty straight-forward change, so I made a commit to address that issue also.

The second behavior was happening because the FilterProvider, which provides FilterContext, was only providing context to the 'Find Properties' page components.  So when you navigated to one of the other pages, anything stored in state in FilterContext was lost.  I've moved FilterProvider into a new FilterProviderWrapper component, modelled after CookieProviderWrapper, and added that to RootLayout.  Now you can navigate to another section of the website, and when you go back to 'Find Properties' the filters are still active.

src/components/SidePanelControlBar.tsx
```diff
@@ -32,11 +32,19 @@ const SearchBarComponent: FC<SidePanelControlBarProps> = ({
...
-  let filterCount = Object.keys(appFilter).length;
-
-  if (shouldFilterSavedProperties) {
-    // Exclude opa_id from filterCount, which counts opa_id as a filter by default
-    filterCount--;
+  const filterCount = () => {
+    let count = 0
+    for (let property of Object.keys(appFilter)) {
+      if (property === "access_process") {
+        count = count + appFilter[property].values.length
```
CodeWritingCow, line 39: Can we shorten this to `count += appFilter[property].values.length`?

src/components/SidePanelControlBar.tsx
```diff
@@ -32,11 +32,19 @@ const SearchBarComponent: FC<SidePanelControlBarProps> = ({
   const savedRef = useRef<HTMLButtonElement | null>(null);
   const { dispatch, appFilter } = useFilter();
 
-  let filterCount = Object.keys(appFilter).length;
-
-  if (shouldFilterSavedProperties) {
-    // Exclude opa_id from filterCount, which counts opa_id as a filter by default
-    filterCount--;
+  const filterCount = () => {
```
CodeWritingCow, line 35: With the rewritten `filterCount`, the UI is displaying the correct count now. However, `filterCount` is now a function, and it's being called multiple times whenever the filter screen loads, or when a user clicks one of the filter buttons.

CodeWritingCow (changes requested):
Hi @jgrimes86 I ran your updated PR locally. You fixed both the first and second issues, and gave a detailed explanation of the second issue. Thanks for going above and beyond.

While the function `filterCount` fixes the filter count issue, it's getting called multiple times. (See screen recording; I added a `console.log` for each time that the UI calls `filterCount`). It'd be more performant if we call it only once or as needed. Otherwise, the rest of the PR looks good.

Can you update the PR so the UI calls `filterCount` only as needed?

https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/d2d88b68-6004-45de-9b0b-18e2b46f3b2c

jgrimes86:
@CodeWritingCow this most recent commit adds a useMemo to filterCount.  I have the useMemo looking for changes to appFilter because filterCount needs to reflect changes to appFilter.  Unfortunately, appFilter "changes" twice every time a filter is applied (although I can see no difference between those changes), which means that filterCount still gets called several times per filter toggle.  But this revision means filterCount isn't called when you first go to the filter view, and I think it halves the amount of times filterCount is called when a filter is toggled.

Video demo of console logs:
https://github.com/CodeForPhilly/clean-and-green-philly/assets/140654473/2f5f971f-bcea-4eab-bca4-8a79ea5ccdbf

CodeWritingCow:
@jgrimes86 I saw your screen recording. The change looks good. I'll go ahead and approve this PR. 👍

CodeWritingCow (approved):

Issue #701, opened 2024-06-21
Bug: Cookie banner reappears after user accepts cookies


CodeWritingCow:
**Describe the bug**
After a user accepts cookies on the Clean & Green website, the cookie consent banner still appears whenever the user opens the site in a new browser window or tab.

**To Reproduce**
1. Go to https://www.cleanandgreenphilly.org.
2. Click on 'Accept' button in the cookie consent banner at the bottom of the screen.
3. Open a new browser window or tab, then go to https://www.cleanandgreenphilly.org.
4. Cookie banner will appear at the bottom of the screen in the new tab/window.

**Expected behavior**
If a user accepts having cookies on the website, and opens the site in a new browser window or tab, the cookie consent banner should not reappear.

**Screenshots**
Here is a screen recording of the bug:

https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/3bf0fd88-e7c0-4da3-9bb4-a45fe488afa7

**Desktop**
 - OS: macOS
 - Browser: Google Chrome and Safari
 - Version: Chrome 126.0.6478.62, Safari 17.5

**Additional context**
This bug is related to Pull Request #635.

JT0Y:
Greetings and salutations-
Could I take a whack at this and get this assigned to me? I was looking through the code and it looks like we aren't using local storage here. I have some test code that I can cleanup and further test if we are okay with that implementation.

CodeWritingCow:
@JT0Y it's yours!

PR #704, opened 2024-06-23
Issue 701: cookie banner re-rendering

JT0Y:
#701 

This resolves the cookie banner re-rendering by recording/checking the user's local storage.

The issue can be recreated by: 
- running localhost
- take action on cookie banner, either accept or decline
- refresh tab, or open new tab and the cookie banner should re-render

Screenshot 2024-06-22 at 7 02 37 PM
Screenshot 2024-06-22 at 7 02 47 PM

### DIFF
--- a/src/components/CookieConsentBanner.tsx
+++ b/src/components/CookieConsentBanner.tsx
@@ -3,15 +3,25 @@
 import { ThemeButton } from "./ThemeButton";
 import { Check, X } from "@phosphor-icons/react";
 import { useCookieContext } from "@/context/CookieContext";
+import { useEffect, useState } from "react";
 
 const CookieConsentBanner = () => {
   const { shouldShowBanner, setShouldAllowCookies, setShouldShowBanner } =
     useCookieContext();
 
+  const [isClientSide, setIsClientSide] = useState(false);
+
+  useEffect(() => {
+    // Ensure accessing localStorage only runs on the client side.
+    setIsClientSide(true);
+  }, []);
+
   const onClickButton = (shouldSaveCookies: boolean) => {
     setShouldAllowCookies(shouldSaveCookies);
     setShouldShowBanner(false);
   };
+
+  if (!isClientSide) return null;
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


CodeWritingCow (changes requested):
Hi @JT0Y, thanks for contributing! I tested your code in my local environment. When I accept cookies and then open the site in a new browser tab, the cookie banner no longer appears.

However, our GitHub repo's Frontend PR Checks are failing due to this error: `ReferenceError: localStorage is not defined`. I'm getting the same error in my localhost ... See screenshot below:

Screen Shot 2024-06-23 at 12 00 31 AM

Can you update the PR to resolve this? Thanks!

JT0Y:
> However, our GitHub repo's Frontend PR Checks are failing due to this error: `ReferenceError: localStorage is not defined`. I'm getting the same error in my localhost ... See screenshot below:
> Can you update the PR to resolve this? Thanks!

yes, I came across this issue and was a bit confused on the issue. I am not familiar with SSR and this environment. But I have additional checks that check if this code is running on a client or not. I did some quick googling on this and this seems to be a well documented issue. Please let me know if this is an unusual pattern, as I am new to next.js & SSR, this _does_ feel strange to me. 

I have tested this and am still able to dismiss the banner and not receiving this issue as well:
```
app-index.js:35 Warning: Prop `className` did not match. Server: "md:flex fixed inset-x-0 bottom-0 z-20 justify-between bg-blue-200 p-4 sm:p-6" Client: "hidden"
```

CodeWritingCow:
@JT0Y I ran the updated PR locally, and the `localStorage` error is no longer appearing.

Your solution looks sound. By checking if `window.localStorage` exists, we can confirm whether the code is running on the client side. Also, React `useEffect` hooks run only on the client side. So when we execute localStorage code inside those hooks, we'll avoid those `undefined` errors. (I'm also new to SSR, so I had to read up a bit on it too.)

Thanks for your work!

CodeWritingCow (approved):

JT0Y:
@CodeWritingCow No worries and thanks for double checking!
Do I need to do anything from here? My understanding from the contributing doc is that I just need to wait for a codeowner to merge it into staging and then close this PR?

CodeWritingCow:
@JT0Y I'll take it from here. Will merge and close this PR!

Issue #700, opened 2024-06-19
Add custom 404 page


CodeWritingCow:
## Description

When users try to visit a nonexistent page, they see a generic 404 error page from Next.js.

Create a custom 404 page to replace the generic page. It should have a link that takes users back to Clean and Green's homepage.

Use @thansidwell's design: https://www.figma.com/design/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?node-id=9209-6244&t=xCAsMNfdaqx6ikBU-1

## Screenshots

Current generic Next.js 404 page:

![Image](https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/5b22322e-4328-455b-b4c6-e12ebec5a5be)

Figma design for new 404 page:

![Screen Shot 2024-06-22 at 6 27 45 PM](https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/862e39fd-83d8-41b6-ac18-52b613818179)

AZBL:
Hi, I'd be interested in this. Would it make sense for me to just do this and come up with a basic design and functionality, or should I wait until there's a design prototype for this?

CodeWritingCow:
Hi @AZBL , thanks for taking the initiative and reaching out! This ticket hasn't been groomed or refined yet. I'd like to get input from the team (especially Nathan from a UX perspective) before marking this ticket as ready for development. So let's hold off on this until Tuesday's standup. Once it's ready, you'll be the first in line for it.

CodeWritingCow:
@AZBL meanwhile I think @KenMain5 is working on a bunch of front-end tickets. If Kenny needs any help, maybe you guys can chat on Slack and see if he has a ticket that might interest you?

CodeWritingCow:
Hi @nlebovits and @thansidwell, whenever users try to visit a nonexistent page, our website currently displays a generic "404 this page could not be found." (See screenshot in ticket description.) Should we make a custom page with a link to the homepage so the user won't be "lost in the wild"?

thansidwell:
@CodeWritingCow Here's a design for the 404 page! Do you think that works?  https://www.figma.com/design/NAFkgq34abW6uJ0R7PW24T/Prototype---Clean-%26-Green-Philly?node-id=9209-6244&t=xCAsMNfdaqx6ikBU-1

CodeWritingCow:
[...]

PR #711, opened 2024-06-25
Issue 700: Add custom 404 page

AZBL:
This adds a custom 404 page.
It adds not-found.tsx to src/app/.
The navbar will still be present.
It has a return to home link. 
It should work for any incorrect url.

clean and green 404

Addresses ticket #700


CodeWritingCow (approved):
Hello @AZBL I tested the PR locally. Custom 404 page appears for wrong URLs. Header is also present. Thanks Alex!

Issue #665, opened 2024-05-31
Make "Before" & "After" text on image an actual text


marvieqa:
## Description

The text "Before" & "After"  on an image found in the "Transform a Property" page is part of the image. We should make this text an actual text element so that users who are using assistive tools can adjust according to the desired visual presentation, such as changing font size, foreground/background color,  line/character spacing or alignment.

## Recommendation
- Put an actual text "Before" / "After" on top or bottom of the appropriate image.

KenMain5:
Can i get assigned to this?

marvieqa:
@JT0Y thanks for picking this up. One option I could think of is to add a caption at the bottom of the photo that depicts these 2 images, like "an example of a Philadelphia lot before and after cleanup" - substituting the image text. Would this work with our current design?

cc @bacitracin @thansidwell ?

bacitracin:
@marvieqa @JT0Y 

Right now the image has some problems on mobile. For example, "after" is clipped on narrow screens. 

![Screenshot 2024-06-25 at 5 24 32 PM](https://github.com/CodeForPhilly/clean-and-green-philly/assets/7329799/5803a42f-20be-4a3d-b44e-5d1de8fbe029)

I think it makes sense to crop the image to remove the bottom sliver that includes the "before" and "after" text, and add a visible caption that appears either before or after the image. After that the image alt text needs updating. 

Currently the alt text is "A Philadelphia lot before a clean up and the same lot, filled with trees and greenery after a clean up." That would be pretty repetitive once we add the caption. Instead I'd remove the alt text, add an id to the new caption, and have `aria-labelledby` point to the new caption id. See example 1 here - https://www.w3.org/WAI/GL/wiki/Using_aria-labelledby_to_provide_a_text_alternative_for_non-text_content

marvieqa:
@bacitracin 100% agree! Maybe we can incorporate the descriptiveness of the current alt text to the caption: "An example of Philadelphia lot before cleanup, and then filled with trees and greenery after cleanup". Something to this effect.

zhenbright:
Can someone assign me this?

bacitracin:
@zhenbright  Assigned! Marvie  and I have made a few comments, but basically the approach we like is:
1. Crop out the text at the bottom of the image
2. Add a visible caption below the image that says  "An example of a Philadelphia lot before cleanup, and transformed with trees and greenery after cleanup". Add an id to this element.
[...]

PR #721, opened 2024-07-01
Zhenbright/665 before after image

zhenbright:
image

TransformPropertyPage.jsx

This addresses the make "Before" & "After" text on image an actual text issue in #665

The issue was caused by the text was embedded in image. 
This PR resolves the issue by removing text from image and putting a caption "An example of a Philadelphia lot before cleanup, and transformed with trees and greenery after cleanup" on bottom of image. 
And removed alt attribute from image tag and used aria-labelledby instead of that.


CodeWritingCow:
hi @zhenbright, the photo and its caption have a noticeable gap:

Screen Shot 2024-07-01 at 5 24 44 PM

CodeWritingCow:
From a design perspective, since the two elements are closely related, visually they should be closer together. Something like this:

Screen Shot 2024-07-01 at 5 24 58 PM

Can we bring them closer together? Thanks!

src/app/(content-pages)/transform-property/TransformPropertyPage.tsx
```diff
@@ -33,12 +33,18 @@ export default function TransformPropertyPage() {
...
           src={beforeAfter}
-          alt={
-            "A Philadelphia lot before a clean up and the same lot, filled with trees and greenery after a clean up."
-          }
+          alt=""
           placeholder="blur"
           className="w-full overflow-hidden rounded-[20px] aspect-video md:aspect-auto object-cover object-center"
+          aria-labelledby="caption-before-after-image"
         />
+        <h2
```
bacitracin, line 41: The caption doesn't look or act like a heading, so let's just change it to a p element. That might fix your "gap" issue as well.

bacitracin: @zhenbright Heads up in case you missed this comment, let's update this h2 to a p element, and fix the spacing.

CodeWritingCow: Hi @zhenbright can you address the PR feedback? Thanks!

CodeWritingCow:
Closing this PR due to lack of activity after 2 weeks.

Issue #656, opened 2024-05-27
Find Properties: Pagination: Add `aria-current` attribute on page buttons


marvieqa:
## Description

Add `<aria-current="page">` on the pagination buttons to give users the context that it is the currently selected page.

## Acceptance Criteria

- add `<aria-current="page">` to pagination buttons

## Additional Info:

Related ticket: https://github.com/CodeForPhilly/clean-and-green-philly/issues/340

KenMain5:
Can someone assign me this?

CodeWritingCow:
@KenMain5 assigned!

CodeWritingCow:
Hi @KenMain5 any progress on this ticket? On this project, If there is no activity on an issue for a week, we will reassign it so other contributors can work on it.

KenMain5:
Hi @CodeWritingCow I recently finished an issue just a few minutes ago. I believe they are the same issue. Can you verify? 
https://github.com/CodeForPhilly/clean-and-green-philly/pull/702#issue-2366911254

CodeWritingCow:
@KenMain5 thanks for the update. The PR is for issue #660, which looks similar to this issue, so I am not sure whether they're the same or not.

@marvieqa can you take a look and see if the two issues are the same or not? Thank you!

bacitracin:
@KenMain5 @CodeWritingCow  @marvieqa  These are different issues. 660 was for the navigation items, this ticket is for the pagination that appears on the Find Properties page. See image:

![Image](https://github.com/CodeForPhilly/clean-and-green-philly/assets/7329799/3e740d59-00b4-460d-9f6c-db7783406bca)

CodeWritingCow:
@bacitracin thanks for clarifying the requirements. Also, I'm moving this ticket back to "To Do" status so other contributors can pick it up.

jgrimes86:
could I work on this issue?

CodeWritingCow:
@jgrimes86 assigned!

PR #723, opened 2024-07-01
Add aria-current to page buttons

jgrimes86:
This PR closes issue #656 

Adds an 'aria-current' attribute to the pagination page buttons.

To test, use a screen reader and navigate to the pagination buttons at the bottom of the property list view of the 'Find Properties' page.

- On pressing an unselected page button, the screen reader should announce, "Current page, page #".
- If you navigate away from the selected page button and then navigate back to it, the screen reader should announce, "Page # button, current page".


bacitracin (approved):
It sounds good, just FYI that another pagination labeling PR exists and will be merged in. I don't think that this interferes with your work. https://github.com/CodeForPhilly/clean-and-green-philly/pull/726

PR #724, opened 2024-07-01
Fix #318: Renamed "Properties in View" and added item count 

### LINKED ISSUE #318: A11y: Consider renaming " x Properties in view" to make it more accurate
## Description

The label ` X Properties in View` doesn't accurately describe what is shown on the page.  We don't actually show all those properties in the current view as they are spread over different pages.  

Also, the current view only shows 6 properties per page. It will be helpful to communicate that as well.

## Recommendation
- Consider renaming it to `X Total Properties` instead.

  <details>
  <summary> Screenshot </summary>

  Screenshot 2024-02-07 at 19 04 39

  </details>

- Add some sort of a counter at the bottom of the page to say `1-6 of 100` (close to the note that says we only display 100)

  <details>
  <summary> sample </summary>

  Screenshot 2024-02-07 at 19 59 23

  </details>

amyyeung17:
## Description
This PR addresses issue #318. Changed "Properties In View" to "Total Properties" and added item count.

### DIFF
--- a/src/components/PropertyDetailSection.tsx
+++ b/src/components/PropertyDetailSection.tsx
@@ -250,6 +250,9 @@ const PropertyDetailSection: FC<PropertyDetailSectionProps> = ({
                 disableCursorAnimation={true}
               ></Pagination>
             </div>
+            <p className="text-center mt-4">
+              {`${((page - 1) * 6) + 1}-${page === pages ? featuresInView.length : (page * 6)} of ${featuresInView.length}`}
+            </p>
             <div className="flex w-full justify-center py-4 px-6">
               <p className="body-sm text-gray-500">
                 Note: only the first 100 properties can be viewed in list.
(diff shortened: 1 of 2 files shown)


src/components/PropertyDetailSection.tsx
```diff
@@ -250,6 +250,9 @@ const PropertyDetailSection: FC<PropertyDetailSectionProps> = ({
                 disableCursorAnimation={true}
               ></Pagination>
             </div>
+            <p className="text-center mt-4">
+              {`${((page - 1) * 6) + 1}-${page === pages ? featuresInView.length : (page * 6)} of ${featuresInView.length}`}
```
bacitracin, line 254: Can we change the "-" to the text "to" instead?

Right now on VoiceOver it reads out as "one six of one hundred" , with the text swap it will be "one to six of one hundred"

amyyeung17: Yep!

bacitracin (approved):
👍 Thanks for updating!

PR #725, opened 2024-07-01
Fix #657: Added aria-live to find properties results 

### LINKED ISSUE #657: Pagination: Previous/Next buttons don't announce the current page
## Description

When a screen reader user selects the Previous or Next button, it doesn't announce anything. Users have no context of what happened or what page they are in. We should have a mechanism to announce the current page when Previous or Next button is selected.

## Recommendation

- Add an `aria-live` region to announce the current page the Next of Previous button took them to.

amyyeung17:
## Description
This PR addresses issue #657. Added aria-live element that announces "You are on page #" after updating pagination.

### DIFF
--- a/data/Dockerfile
+++ b/data/Dockerfile
@@ -4,8 +4,7 @@ FROM python:3.11.4
 # Set the working directory in the container
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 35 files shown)


amyyeung17:
Whoops, I didn't mean to merge into main.

Issue #650, opened 2024-05-26
Methodology Page: Screen reader focus doesn't land on the intended heading.


marvieqa:
## Description

The "Learn more" links from the Filter UI point to span IDs in the Methodology page, where  upon navigating to it, the topic is framed beautifully in view.  However,  when using a screen reader this gets confusing. Here's an example:

Get Access "Learn More link" points to `/methodology/#access-method`
  <details>
  <summary>screenshot</summary>

  <img width="1223" alt="Screenshot 2024-05-26 at 23 08 31" src="https://github.com/CodeForPhilly/clean-and-green-philly/assets/102739972/15fc88c7-aedd-4634-ac96-abe88a9d6aa0">

  </details>

In Methodology page, this ID is pointing in this `<span>`, I believe because we wanted the "How did we determine “priority”?" heading to be centered in view:
  <details>
  <summary>Screenshot</summary>

  <img width="1272" alt="Screenshot 2024-05-26 at 23 12 35" src="https://github.com/CodeForPhilly/clean-and-green-philly/assets/102739972/50cf771c-1103-4de9-83d9-cbcd917af6bc">

  </details>

However, this is putting the screen reader focus on the paragraph above the intended heading, thus screen reader announces this first. 
A screen reader user may get thrown off when this paragraph is announced without a title or context, because coming from the Filter UI link we tell them this link goes to Methodology - Priority Level. The expectation would be that it announces the heading for Priority Level topic.

## Recommendation

- Put the span before the heading (or put the ID on the heading itself and test if it works), even if it means the heading will not be beautifully centered in view, so that screen reader announces the heading right away.

jgrimes86:
Can I be assigned this one?

PR #729, opened 2024-07-02
Moved "learn more" link fragment spans to be above subject header

jgrimes86:
This PR closes issue #650 

I moved the `<span>`s that serve as targets of the two "learn more" links so that they are immediately above the headers of the sections they are related to.  Visually, the linked-to headers are brought to the top of the Methodology page, and screen readers now announce the header, rather than the previous paragraph.

Question for @bacitracin :  I'm still using the `<span>`s so that the screen reader reads the text of the header and then stops.  I could delete the `<span>`s and instead add `id=...` to the `<div>` containing the header and paragraph text, which was suggested as a possible alternative solution to this issue.  I tested it and the navigation works the same either way.  However, if I add the `id` to the `<div>`, the screen reader reads the header and also reads the first paragraph.  What is the preferred behavior?

Also, for reference, here are some before-and-after screenshots of the "access method" section of the Methodology page as it appears after clicking the "learn more" link:
Before this PR on a large screen:
Access link large screen original
After this PR on a large screen:
Access link large screen revised

Before this PR on a small screen:
Access link small screen original
After this PR on a small screen:
Access link small screen revised

### DIFF
--- a/src/app/(content-pages)/methodology/MethodologyPage.tsx
+++ b/src/app/(content-pages)/methodology/MethodologyPage.tsx
@@ -25,8 +25,6 @@ export default function MethodologyPage() {
           residents, CDCs, City government offices, academic researchers, and
           more.
         </p>
-        {/* Offset for clicking on learn more from filter, scrolls to section below */}
-        <span id="priority-method"></span>
         <p className="body-md">
           Although we aim to simplify the decision-making process for users of
           Clean & Green Philly, we value transparency. Below, we lay out key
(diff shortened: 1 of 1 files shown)


bacitracin:
@jgrimes86 Thanks for the screenshots! They're helpful. It looks good to me.

I think in the ticket @marvieqa meant moving the id to the H2 that says "How did we determine “priority”?" and not the container div. That would be cleaner if it works (haven't tried but don't see any issues with it).

It's up to you whether you want to try that or keep the span. We definitely don't want the entire paragraph read out, that's a rough experience for screenreader users.

jgrimes86:
> I think in the ticket @marvieqa meant moving the id to the H2 that says "How did we determine “priority”?" and not the container div. That would be cleaner if it works (haven't tried but don't see any issues with it).

Ah, that makes more sense.  I made that change and it seems to behave the way we want it.  Navigation works the same as I had it before, and the screen reader just reads the H2 text.

bacitracin (approved):
Less code, nice!

Issue #712, opened 2024-06-25
Task: Show error message if map fails to load


CodeWritingCow:
## Describe the task
When the map fails to fetch a `pmtiles` file from Google Cloud, hide loading spinner and show an error message in `src/components/PropertyMap.tsx`

Currently, the map page shows the loading spinner spins endlessly with the message "Loading properties." This makes for a poor user experience.

## Acceptance Criteria

- [x] Hide loading spinner
- [x] Show error message instead of "Loading properties"
- [x] Show error message saying, "We are having technical issues. Please try again later."

## Additional context

* This ticket is related to #709. While working on that PR, I noticed that if the map page got a 404 error for the `pmtiles` file, it would get stuck on "Loading properties." The loader would spin endlessly.
* Here's a screenshot of what that looks like to a user:

<img width="1274" alt="Screen Shot 2024-06-25 at 11 41 15 AM" src="https://github.com/CodeForPhilly/clean-and-green-philly/assets/7751862/02b1994c-7d09-433f-8f1b-1fc9a494a76e">

Michael-Dratch:
Can I pick up this task?

PR #732, opened 2024-07-06
Michael dratch/712 error message if map fails to load

Michael-Dratch:
This PR is for this task: Show error message if map fails to load #712

https://github.com/CodeForPhilly/clean-and-green-philly/issues/712

To view the error state go to the find properties page, open dev tools, and block the URL from google storage to simulate a failed API call.

image

On page reload, this error message will be displayed on the property details panel.

image

I think the automatic prettier formatting makes it seem like a lot of lines of code have been changed. But most of these lines are minor changes in formatting. If it would be better for the PR I can redo do these changes with out any of the formatting changes and resubmit.

### DIFF
--- a/src/app/find-properties/[[...opa_id]]/page.tsx
+++ b/src/app/find-properties/[[...opa_id]]/page.tsx
@@ -33,7 +33,7 @@ const MapPage = ({ params }: MapPageProps) => {
   const [featureCount, setFeatureCount] = useState<number>(0);
   const [currentView, setCurrentView] = useState<BarClickOptions>("detail");
   const [loading, setLoading] = useState(true);
-  const [loadingError, setLoadingError] = useState(false);
+  const [hasLoadingError, setHasLoadingError] = useState(false);
   const [selectedProperty, setSelectedProperty] =
     useState<MapGeoJSONFeature | null>(null);
   const [isStreetViewModalOpen, setIsStreetViewModalOpen] =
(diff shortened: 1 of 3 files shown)


src/app/find-properties/[[...opa_id]]/page.tsx
```diff
@@ -33,6 +33,7 @@ const MapPage = ({ params }: MapPageProps) => {
   const [featureCount, setFeatureCount] = useState<number>(0);
   const [currentView, setCurrentView] = useState<BarClickOptions>("detail");
   const [loading, setLoading] = useState(true);
+  const [loadingError, setLoadingError] = useState(false);
```
Comment (CodeWritingCow) [inline src/app/find-properties/[[...opa_id]]/page.tsx:36]: Can we rename `loadingError` and its setter to something like `hasLoadingError` and `setHasLoadingError`?

This variable has a boolean value, but its current name can give the wrong impression. `loadingError` sounds like a name for something else, such as an error response object.

Comment (CodeWritingCow) [reply src/app/find-properties/[[...opa_id]]/page.tsx:36]: This variable has a boolean value, but its current name can give the wrong impression. `loadingError` sounds like a name for something else, such as an error response object. (I realize the line above has a boolean variable named `const loading`, but let's not follow that example.)

src/components/PropertyDetailSection.tsx
```diff
@@ -50,6 +50,7 @@ interface PropertyDetailSectionProps {
   featuresInView: MapGeoJSONFeature[];
   display: "detail" | "list";
   loading: boolean;
+  loadingError: boolean;
```
CodeWritingCow, line 53: Please see previous comment about renaming the variable.

CodeWritingCow (changes requested):
Hi @Michael-Dratch I tested your PR. The error message appears correctly if the map API call fails. Once you make the changes mentioned in my comments, I'll approve and merge your PR. Thanks!

Michael-Dratch:
Hi @CodeWritingCow, thanks for the feedback. I just made those requested updates and pushed the changes.

CodeWritingCow (approved):
@Michael-Dratch updated PR looks good. Thanks for your contribution!

PR #739, opened 2024-07-09
Issue 645: schedule backend job on GCP

zigouras:
This PR contains fixes, enhancements and documentation for the scheduled production cron job on the new GCP virtual machine.


data/src/config/config.py
```diff
@@ -33,9 +33,16 @@
 tiles_file_id_prefix: str = "vacant_properties_tiles"
 """ the prefix of the name of the tiles file generated and saved to GCP """
 
+write_production_tiles_file: bool = False
```
CodeWritingCow, line 36: Can we rename `write_production_tiles_file` to something like `should_write_production_tiles_file`?

This variable has a boolean value, but its name sounds like it could be a function. Let's make its meaning and purpose more clear.

data/src/config/config.py
```diff
@@ -33,9 +33,16 @@
...
+""" Whether to write the main vacant_properties_tiles.pmtiles as well as the staging vacant_properties_tiles_staging.pmtiles. 
+BE CAREFUL, if true this writes the production file.
```
CodeWritingCow, line 38: Good warning

CodeWritingCow (approved):
@zigouras this PR looks good to me

CodeWritingCow:
@nlebovits can you take a look at this PR in case I missed anything? Thanks!

Issue #638, opened 2024-05-21
DevOps: Add pr_checks_backend 


zackdreitzler:
## Describe the task
Add the pr_checks_backend.yml file as a GitHub Actions workflow. The goal of this is to show every developer/reviewer that a PR impacting the `data` folder can successfully build and run the `vacant-lots-proj`. This also sets a foundational workflow that can be used to run a data quality check, formatter, and/or linter checks.

## Acceptance Criteria

- [ ] The workflow should only trigger when changes are found in the `data` directory or any of the subdirectories
- [ ] The workflow should contain a `docker-compose run vacant-lots-proj` step
  - [ ] On success this step reports success to the PR that triggered this action
  - [ ] On failure this step reports a failure to the PR that triggered this action

## Additional context

This is currently blocked by #630. The postgres DB is required to get the docker build and run to be successful.

nlebovits:
closed by #756

Issue #640, opened 2024-05-21
DevOps: Add linting step to pr_checks_backend workflow


zackdreitzler:
## Describe the task
Add a linter check to any PR impacting the `data` directory and its sub directories. This check ensures that new python code is passing basic linting checks.

## Acceptance Criteria

- [ ] A linter checking step exists on the pr_checks_backend workflow
- [ ] When no issues are found, report a success to the PR that triggered this action
- [ ] When one or more issues are found, report a failure to the PR that triggered this action

## Additional context

This story is currently blocked until #638 is completed. This story is also blocked pending a decision on what python linter to use.

nlebovits:
closed by #756

PR #756, opened 2024-07-16
get pr_checks_backend.yml working

nlebovits:
Addresses #638 . Adds a test to confirm that the primary docker container builds correctly when any files are changed in `data`. Probably would be good to also add checks for the `postgres` container.

### DIFF
--- a/.github/workflows/pr_checks_backend.yml
+++ b/.github/workflows/pr_checks_backend.yml
@@ -31,13 +31,15 @@ jobs:
 
       - name: Run Black Formatter in Docker
         run: |
+          cd data
           docker-compose run --rm formatter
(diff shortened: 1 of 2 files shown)


nlebovits:
should now also address #640

Michael-Dratch:
I tried running this workflow on a test branch and got this error on the Black formatter check step:

image

I'm pretty new to git-hub workflows but I think you might need to cd into the directory with the docker-compose file before running the docker-compose command.

CodeWritingCow:
> Hi @nlebovits
> 
> I tried running this workflow on a test branch and got this error on the Black formatter check step:
> 

@Michael-Dratch thanks for catching this!

Issue #785, opened 2024-07-18
DevOps: update dependabot to open PRs against staging, limit to 1 PR per ecosystem


nlebovits:
## Describe the task
Update dependabot to open PRs against staging rather than main, and combine all PRs into a single PR per ecosytem (ghactions, npm, pip). 

Code:

```sh
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "daily"
    target-branch: "staging"
    open-pull-requests-limit: 1
    commit-message:
      prefix: "deps"

  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "weekly"
    target-branch: "staging"
    open-pull-requests-limit: 1
    commit-message:
      prefix: "deps"

  - package-ecosystem: "pip"
    directory: "/data/src"
    schedule:
      interval: "weekly"
    target-branch: "staging"
    open-pull-requests-limit: 1
    commit-message:
      prefix: "deps"
```

PR #786, opened 2024-07-18
update dependabot to open all PRs against staging rather than main

nlebovits:
limit dependabot PRs to 1 per package ecosystem to minimize excessive PRs


nlebovits:
closes #785

PR #880, opened 2024-08-20
Update Contributor Section in READ.ME

mjnishigaki:
Updating the Contributor section in the READ.ME file to where the title is more accurate to being a Point of Contact section for new users.


zigouras:
I think 'Points of Contact' is more accurate.

PR #921, opened 2024-09-29
fix postgis version, update backend docs

zigouras:
@jinks145 found that when trying to build the postgres container, the postgis version in Dockerfile-pg was no longer supported.   I reproduced the error:
 E: Version '3.4.3+dfsg-1.pgdg110+1' for 'postgresql-16-postgis-3' was not found

I updated the version to 3.4.3+dfsg-2.pgdg110+1 per his suggestion which allowed the build to complete.

I also made some small tweaks to the backend setup doc.

### DIFF
--- a/data/Dockerfile-pg
+++ b/data/Dockerfile-pg
@@ -5,11 +5,11 @@
 FROM postgres:16-bullseye
 
 LABEL maintainer="PostGIS Project - https://postgis.net" \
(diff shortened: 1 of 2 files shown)


zigouras:
> @zigouras I think when you used `cmd` + `f` to replace `docker-compose` with `docker compose`, you also modified the filename for the `docker-compose.yml` in the backend docs. Can you correct?

No, I added the hyphen back into the file name where it was missing before.

PR #933, opened 2024-10-02
Replaced logo.svg with minified version

AZBL:
Reduces file size of logo.svg from 9447 bytes to 5547 bytes. 

@CodeWritingCow


CodeWritingCow (approved):
Thanks @AZBL!

PR #956, opened 2024-10-17
amberroseweeks-917-data-disclaimer

Amberroseweeks:
This resolves issue https://github.com/CodeForPhilly/clean-and-green-philly/issues/917

- modal closes when user presses X, ESC, or the "I understand" button
- Currently this loads whenever the map page loads, this could get annoying. I wonder if we should include a way to only load for the first time a user loads the page? I have to look into that more.
- Using design specs from https://www.figma.com/design/XKz7lqtmj4cvxTDMZWrEmm/Disclaimer-Overlay---Updated?node-id=1-1206&node-type=instance&t=da7aVMN5rGxhouML-0 but I adjusted the corners and the padding.

Screen Shot 2024-10-17 at 3 09 55 PM


nlebovits:
@Amberroseweeks could you use local cookies to have it only open the first time a user lands there? That's how we handle our consent banner: https://github.com/CodeForPhilly/clean-and-green-philly/blob/main/src/components/CookieConsentBanner.tsx

zigouras:
I would suggest having a static piece of text on the page that just stays there.  It is a useful piece of information for every visit to the map and having popups, like cookie consent dialogs, is intrusive.

nlebovits:
@zigouras I think for the moment, intrusive is what we want. This is a serious issue in the city's data, and we don't have even a short-term fix in place. We can switch to something less obtrusive when we have a real fix in place. And I also want people to know that it's the City's fault, not ours.

CodeWritingCow (approved):
@Amberroseweeks banner looks good, and it works on desktop and mobile.

CodeWritingCow:
If we need to get the disclaimer up ASAP, then I can merge this PR, and @Amberroseweeks can submit a new PR later to make it appear only for first-time page load via local cookie. If there's no rush on this, then we can wait until she incorporates local cookie.

nlebovits:
@CodeWritingCow how long would local cookie take? if it's just a day or two, we can wait

CodeWritingCow:
@nlebovits It should take less than a day.

nlebovits:
@CodeWritingCow let's wait until @Amberroseweeks has cookies integrated, then

@zigouras I'll ask the UX team to mock a less obtrusive indicator for the last data update, too--that can replace the dislclaimer once I get a patch in for the vacancy data, which should be within the next week or two

Amberroseweeks:
Okay thanks all I will work on incorporating the cookies and get back to you soon!

Amberroseweeks:
@nlebovits @CodeWritingCow can you review the updates? Thank you!

nlebovits:
@Amberroseweeks looks good to me! thanks for the quick turnaround. @CodeWritingCow feel free to merge when you've signed off on it

CodeWritingCow (approved):
@Amberroseweeks disclaimer banner works as expected. It appears on initial page load and doesn't appear later. Thanks!

PR #960, opened 2024-10-18
Lebovits/issu946 add backup vacant land data

nlebovits:
This PR adds a check to the pipeline to confirm that the vacant lots dataset contains at least 20,000 observations. If it does not, it defaults to the June 2024 vacant lots dataset provided by CityGeo (the last known accurate version of the dataset) and loads it from our GCS bucket. 

Also a small fix: I added the Google Cloud Project explicitly to the Docker env.


nlebovits:
Going to test the full pipeline locally to confirm everything looks okay and then will merge.

Issue #848, opened 2024-07-30
Report any errors from the backend script to Slack.


zigouras:
**Is your feature request related to a problem? Please describe.**
This morning the scheduled backend job failed with the error below:
cleanandgreenphl@backend:~/clean-and-green-philly/data$ docker logs  data-vacant-lots-proj-run-bd197eebfaa5 2>&1 | grep -i error
"ValueError: tmp/temp_vacant_properties_tiles_merged.pmtiles is 4557995 bytes in size but should be at least 5242880.  Therefore, we are not uploading any files to the GCP bucket.  The file may be corrupt or incomplete."

If the backend job fails, we will know indirectly by the fact that the diff report will not get sent to the `clean-and-green-philly-back-end` Slack channel.  However, it would be better to have the error directly reported to the Slack channel.

**Describe the solution you'd like**
Modify the Python code of the backend script to send a message to Slack on any exception raised.
This can be done by putting a try/catch block around all of the logic in the `script.py` Python code.  Then use the existing Slack API and credentials to send the error message to the channel.  See existing code in `diff_report.py::send_report_to_slack`

Once the change has been made in the staging branch, do a `git fetch` and `git pull` in the project directory on the GCP VM to deploy the latest code to the scheduled job.

**Describe alternatives you've considered**
If there is a better way to do this, please suggest.

**Additional context**
Supply unit tests with the code change.

nlebovits:
closed by #968

PR #968, opened 2024-10-19
Lebovits/issu848 report backend errors to slack

nlebovits:
Adds a try-catch block to the main script to report any back end errors to the Slack channel. Also includes unit test to assert functionality (both when script runs without error and with).


nlebovits:
closes #848

PR #1004, opened 2024-11-16
Amberroseweeks 625 improve filter button

Amberroseweeks:
This resolves issue https://github.com/CodeForPhilly/clean-and-green-philly/issues/625

The button when activated has aria-expanded when selected. The focus skips the Download button and goes directly to the "X" button in the side panel. 

Screen Shot 2024-11-16 at 4 44 08 PM


bacitracin (approved):
LGTM

PR #1017, opened 2024-11-23
Add timescale and pg_stat_statement postgres extensions

rmartinsen:
Adds timescaledb and pg_stat_statement extensions for postgres. See #997 for detailed discussion. For now, the timescale extension is installed in a separate container with postgres listening on port 5434. This may be consolidated with the existing postgres container in the future.


nlebovits:
@rmartinsen looks good! We can make tweaks as needed as we get the pipeline set up

PR #1064, opened 2024-12-30
Filtering for only main zones

DonovanAndrews300:
Description
This PR updates the zoning filter dropdown to display only main zones as filter options, while automatically applying the corresponding subzones. Additionally, the PR introduces a feature allowing users to search for specific filter options by typing directly into any dropdown. See issue #1058 

Steps to Test
1. Navigate to the property map.
2. Open the filter pane.
3. Check the Zoning filters to verify that only main zones are listed as filter options.
4. Confirm that all main zones are correctly listed.
5. Ensure that selecting a main zone applies all relevant subzones.
6. While focused on any filter dropdown, type to search for specific options.

Screenshots 
[Screencast from 12-29-2024 09:17:00 PM.webm](https://github.com/user-attachments/assets/6ff16358-0ed9-4309-9907-164a4184fdcc)

PR #1071, opened 2025-01-08
Deploy Staging to Main branch

CodeWritingCow:


CodeWritingCow:
Just checked the code in staging. Looks like there's a bug where all properties disappear from the map when user selects a "suggested priority" filter. See screen recording.

https://github.com/user-attachments/assets/a04135d6-225d-4332-a0ac-b1ee23af2932

CodeWritingCow:
I'll look into what might be causing this buggy behavior.

CodeWritingCow:
Retested this PR in Vercel preview env. I confirmed the following:

- New search box works
- "Suggested Priority" filter works again
- Changes to "Zoning" filter reverted (PR #1064 reverted)

See screen recording:

https://github.com/user-attachments/assets/656441a5-a37b-4bc4-a20f-432ddeaab9c7

CodeWritingCow:
We'll revisit PR 1064 later and see if it can be reimplemented without breaking the "Suggested Priority" filter.

Issue #979, opened 2024-10-27
A11y: Mobile nav menu items should be navigable via keyboard up and down keys


bacitracin:
## Describe the task

When the screen is narrower and the mobile nav hamburger button pops up, it is not reachable or operable via keyboard navigation. Since it's not reachable by keyboard, screenreaders are affected as well.

This ticket is the SECOND part of making the mobile nav more accessible.[ This ticket](https://github.com/CodeForPhilly/clean-and-green-philly/issues/978) must be completed first. 

1. You should be able to use UP and DOWN arrow keys to navigate the menu items.
2. Focus state should be visible
3. Pressing ENTER or SPACE will choose the option. The menu should close and you'll go to that page.

## Resources
Total expected behavior of the menu: 
[(First ticket)](https://github.com/CodeForPhilly/clean-and-green-philly/issues/978):
You should be able to TAB to the hamburger menu, press ENTER or SPACE and the dropdown opens. 

(THIS ticket)
Then you can use the UP or DOWN arrow keys to navigate the items (with focus state visible). Press ENTER or SPACE to choose one of the items. The menu should close and you'll go to that page.

[(Third ticket)
](https://github.com/CodeForPhilly/clean-and-green-philly/issues/980)While the dropdown is open pressing ESC or TAB will close the dropdown and focus will go back to the hamburger button. The page will not change.

See this example for most of the desired behavior (except for the behavior while the dropdown is open and pressing TAB.) https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/#mythical-page-content

millmason:
Just making a note for any future implementer, since I'm working on the preceding, related ticket ( #978 ) The mobile nav menu items are currently navigable via tab as expected -- what is _not_ expected is that they are the absolute last items in tab order for the page. If you inspect the HTML of the page, they're at the very bottom of the document (explaining the tab order.) I don't know why this would be, since in the React code, the mobile menu is fully located inside of the `Header` component.

millmason:
@bacitracin I've been looking into this a bit -- not sure what the lift is going to be like, but would you mind if I took it on?

CodeWritingCow:
@millmason assigned! Mill and @bacitracin, #978 just went live, so this ticket should be unblocked.

bacitracin:
@millmason @CodeWritingCow thanks for the heads up, let me know if you have any questions

Issue #978, opened 2024-10-27
A11y: Mobile nav button should be reachable via keyboard


bacitracin:
## Describe the task

When the screen is narrower and the mobile nav hamburger button pops up, it is not reachable or operable via keyboard navigation. Since it's not reachable by keyboard, screenreaders are affected as well.

This ticket is the first part of making the mobile nav more accessible. 

1. You should be able to TAB to the hamburger menu (probably need to add tabIndex=-1), 
2. You should be able to press ENTER or SPACE and the dropdown opens.

## Resources
Total expected behavior of the menu: You should be able to TAB to the hamburger menu, press ENTER or SPACE and the dropdown opens. Then you can use the UP or DOWN arrow keys to navigate the items (with focus state visible). Press ENTER or SPACE to choose one of the items. 

While the dropdown is open pressing ESC or TAB will close the dropdown and focus will go back to the hamburger button. 

See this example for most of the desired behavior (except for the behavior while the dropdown is open and pressing TAB.) https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/#mythical-page-content

millmason:
🌤️ Hello! I believe I have a fix for this. Would you mind assigning it to me? @bacitracin

bacitracin:
Assigned

millmason:
Looking good on staging so far 🤞

millmason:
Looking good on live! I think we're set to close this ticket.

CodeWritingCow:
@millmason I'll close this ticket as complete!

PR #1081, opened 2025-01-15
Add tabIndex to mobile nav toggle

millmason:
## Why
The mobile hamburger menu is not currently accessible either by keyboard or screenreader, per issue #978 . 

## What
As noted by @bacitracin in #978 , the required change is extremely simple. We just add a `tabIndex` onto the mobile nav toggle and the browser takes care of the rest.

## NB
I've added a comment to the ticket which follows this one, #979 , but the menu items for the mobile menu are accessible via tab...but only at the very very end of the tab order, past the bottom of the page. This is because the mobile menu is loaded into the html at the very end of the document, in spite of being located within the `Header` component in the React code. I don't know why this would be! Possibly not an issue since we want to access them via keyboard up-down arrows anyway, but thought it would be worth noting in case anyone has insights. 

### Current behavior in production:
(User tabs through page, starting with main `Clean & Green Philly` icon in nav, and tabs directly into the page content without accessing the `Menu` hamburger toggle.)

https://github.com/user-attachments/assets/f6c1398d-0a25-4043-823e-f1e42a421216

### Adjusted behavior in local with these changes:
(User tabs through page and successfully tabs to `Menu` hamburger toggle, which opens and closes appropriately on `Enter`.)

https://github.com/user-attachments/assets/1bd66665-b1d6-4cdd-a732-293a1542d03a

## Steps to test
- Pull down these changes to your local machine
- Build changes with usual steps
- Observe the page at a width < 850px to see mobile menu
- Tab through the page until you successfully `focus` the mobile 
- Mobile should display appropriate `focus` styles (a standard blue halo around the element)
- On key press `Enter` or `Space` the menu should open
- Hitting the `Enter` or `Space` key again should close the menu

### DIFF
--- a/src/components/MobileNav.tsx
+++ b/src/components/MobileNav.tsx
@@ -45,6 +45,7 @@ const MobileNav: FC = () => {
           as="nav"
           aria-label={isMenuOpen ? 'Close menu' : 'Open menu'}
           className="min-[850px]:hidden flex-end w-fit"
+          tabIndex={0}
           icon={
             <>
               <PiList className="h-6 w-6 linkIcon" /> Menu


CodeWritingCow (approved):
Tested this PR. I was able to reach the mobile nav button by pressing the "TAB" key on my keyboard. Thanks @millmason!

Issue #1079, opened 2025-01-14
Mobile nav - Hover styling does not appear in menu on viewports smaller than 640px


CodeWritingCow:
**Describe the bug**

Mobile nav menu items are not changing background color when user hovers over menu items on screens smaller than 640px.

**To Reproduce**

1. Narrow the width of the screen to smaller than 640px to get the nav to switch to the hamburger icon.
2. Click on the hamburger icon to trigger the nav dropdown to show
3. Use your mouse to hover over the options

**Expected behavior**
On screens smaller than 640px, mobile nav menu background color should change to gray background when user hovers over nav menu items.

**Screenshots**
Mobile nav menu item not changing background color on hover in viewports smaller than 640px.

<img width="626" alt="Screen Shot 2025-01-13 at 8 11 02 PM" src="https://github.com/user-attachments/assets/babbe387-b7c4-4178-841c-ae2fa738e523" />

**Desktop (please complete the following information):**

- OS: MacOS Monterey v12.7.6 
- Browser: Google Chrome v131.0.6778.205

millmason:
@CodeWritingCow I think I have a fix for this! Would you mind assigning it to me?

CodeWritingCow:
@millmason assigned!

millmason:
Looking good on staging so far! 👀

millmason:
🚂 looking good on live! I think we can close this one

CodeWritingCow:
@millmason Thanks for checking it! I'll close this ticket as complete.

PR #1082, opened 2025-01-15
Allow hover styles on screens below 640px

millmason:
## Why
Changes requested in #1079 to unify mobile styles across screenwidths. 

## What
Hover styles for the mobile menu items below the `max-sm` screenwidth (640px) were being overridden by a tailwind rule applied to the `iconLink` css class. Removing that rule restored them.

#### Current behavior in production:
https://github.com/user-attachments/assets/436f2ff8-1924-4e01-8817-c6fbd44c7b33

#### Modified behavior with this change:
https://github.com/user-attachments/assets/abdc9c0e-c6a5-433c-9b00-7acfd27cdf24

### Observe existing behavior
1. Navigate to the prod site at https://www.cleanandgreenphilly.org/
2. Observe the site at a screenwidth of more than 640px, and less than 850px.
3. Click on the mobile nav toggle (the hamburger icon in the nav bar)
4. With the mobile nav open, hover over the nav items. There should be a light gray background behind items.
5. Observe the site at a screenwidth of less than 640px
6. Open the mobile nav again as above, and hover over the menu items
7. There should be no background visible on the menu item while hovering
### Observe modified behavior
1. Build this code locally
2. Navigate to the local site at `localhost:3000` (or whichever port you're using if other)
3. Observe the site at a screenwidth of less than 640px
4. Click on the hamburger icon to open the mobile menu
5. Hover over menu items
6. Menu items should show a light gray background on hover, as observed above at screenwidth > 640px and less than 850px.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -4,8 +4,7 @@
 
 @font-face {
   font-family: 'HK Grotesk';
-  src:
-    url('/fonts/hkgrotesk-bold-webfont.woff2') format('woff2'),
+  src: url('/fonts/hkgrotesk-bold-webfont.woff2') format('woff2'),
     url('/fonts/hkgrotesk-bold-webfont.woff') format('woff');
   font-weight: bold;
   font-style: normal;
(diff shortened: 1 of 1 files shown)


CodeWritingCow (approved):
@millmason I tested your PR and verified that the hover style is correct. Thanks for fixing this!

Issue #1024, opened 2024-11-24
Remove scroll jump when applying filters


thansidwell:
Currently when you apply a filter, it immediately it scrolls up to the top of the filter, making it disorienting for users to use and breaking their concentration. 

https://github.com/user-attachments/assets/2a9d9fe1-c22a-4ba7-b748-1d869c40ac2d

When a user clicks on a filter it should not scroll at all, but stay exactly where it was before, so users can have a smooth experience.

vinhyan:
From my initial investigation, this occurs because the filters are being `map`ped with a unique `key`, causing React to re-render each time a filter option is selected (and its state is updated). As a result, the scroll position resets to the top of the element after every re-render.

I can look more into this and give it a try. @nlebovits

nlebovits:
Thank you! I'll assign you now.

Btw, we'll have high fidelity mocks for the contact page by this weekend. Haven't forgotten that.

wakoong:
I would like to take a look at this one as well!

CodeWritingCow:
@wakoong assigned!

PR #1097, opened 2025-01-31
Bugfix - prevent losing focus when filtering

wakoong:
### Description
This PR fixes a bug, #1024, where clicking on filtering options from the Find Properties page causes the focus to move to the top.

### Cause of the Bug
Not having currentView in the dependency array caused the useEffect to run on every render. Adding currentView as a dependency ensures that the re-render is triggered only when the currentView state changes.

### Steps to Test
1. Navigate to 'Find Properties'
2. Click on the 'Filter' button
3. If you click on any filtering action on the page, the focus should remain at the point where the action occurred.

### Screen Recording

https://github.com/user-attachments/assets/2290409e-a431-4340-bf73-c59cf94a26ab

### DIFF
--- a/src/components/SidePanelControlBar.tsx
+++ b/src/components/SidePanelControlBar.tsx
@@ -80,7 +80,7 @@ const SearchBarComponent: FC<SidePanelControlBarProps> = ({
         closeButton.focus();
       }
     }
-  });
+  }, [currentView]);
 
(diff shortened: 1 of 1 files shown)


CodeWritingCow (approved):
@wakoong bugfix worked correctly. Thank you!

Issue #1038, opened 2024-12-05
Task: Upgrade `cross-spawn` version due to Node Cross-Spawn Vulnerability (CVE-2024–21538)


vinhyan:
## Describe the task

Looks like the issue of `cross-spawn` might be from this transitive dependency, upgrading the parent dependency is needed:
![image](https://github.com/user-attachments/assets/8faff8ed-974d-4ddc-b254-cc89ccccef51)

High vulnerability issue with the current `cross-spawn` version 7.0.3. [CVE-2024-21538
](https://avd.aquasec.com/nvd/2024/cve-2024-21538/)

`npm audit fix` cannot fix unless bumping up version to 7.0.5 or above.

## Acceptance Criteria

- [ ] `cross-spawn` dependency is updated to version 7.0.5 or higher in the npm package.
- [ ] The vulnerability CVE-2024-21538 in `cross-spawn` is no longer flagged by `npm audit`.
- [ ] The fix does not break any existing functionality or dependencies in the project.

## Additional context

- This issue is affecting code commits due to a failure in `npm audit`.
![CVE-2024-21538](https://github.com/user-attachments/assets/ec0028b3-b240-4fbd-9292-6e325d1f3afe)

vinhyan:
Our locked `npm` version is 10.8.3, which uses `cross-spawn` v7.0.3—a version with the above vuln. I tested modifying the `npm` version to 10.9.1 directly in the lockfile, deleted `node_modules`, and ran `npm i`. This approach worked, as `npm` 10.9.1 uses `cross-spawn` 7.0.6, which does not have the vuln issue. However, I’m not sure if manually editing the lockfile is recommended.

This issue is currently blocking me from committing code, so any advice on resolving it would be greatly appreciated.  :) @CodeWritingCow

vinhyan:
@nlebovits also looping you in for advice on this. Thanks! :)

nlebovits:
Hey @vinhyan sorry for my slow response on this! Was OOO while traveling. I'm not a JS expert at all but I'll make sure @CodeWritingCow sees this and gets back to you.

CodeWritingCow:
@vinhyan When I ran `npm update` and then `npm audit` locally, the vulnerability alert for cross-spawn disappeared. 

Also ran `npm ls cross-spawn` to verify that cross-spawn got upgraded to v7.0.6:
[...]

PR #1110, opened 2025-02-20
Cross-spawn Vulnerability Fix and Other Audit Fixes

cfreedman:
## Description

Update the lockfile to fix the security vulnerability affecting `cross-spawn` versions `7.0.0 - 7.0.4`. The flagging module was a transitive dependency of `cross-spawn` being built within the `npm` module, so `npm update npm` was enough to fix the issue, but I also included an override component in the `package.json` to prevent that module from being rebuilt in the event of any issue with needing to generate the lockfile.

This PR also includes audit fixes to the lockfile from several other medium level security issues that were outstanding and flagged by the precommit hook.

## Related Issue(s)

This PR addresses issue #1038 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

I pulled a fresh repo and `npm update npm` , `npm audit fix`, and then `npm install` to build the project modules, and this was enough to push the versions of `cross-spawn` appearing the repo to the appropriate, safe versions. This can be verified with `npm ls cross-spawn`.

I also deleted the `node_modules` and `package-lock.json` and rebuilt both using only the `package.json` with the override statements and this also ensured the issue would be fixed if rebuilding those portion from scratch were ever needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "1.0.0",
       "dependencies": {
         "@heroicons/react": "^2.1.5",
-        "@maptiler/sdk": "^2.5.1",
         "@maptiler/geocoding-control": "^1.4.1",
+        "@maptiler/sdk": "^2.5.1",
         "@nextui-org/react": "^2.4.6",
         "@phosphor-icons/react": "^2.1.7",
         "@turf/centroid": "^7.0.0",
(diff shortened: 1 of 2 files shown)


CodeWritingCow (approved):
@cfreedman thank you!

Issue #1094, opened 2025-01-28
Task: Implement ability to key in neighborhoods and community organizations in property search


tcmoorman:
## Describe the task

During workshops and interviews following V1 launch, users expressed frustration with having to scroll to find their neighborhood and community organization from the list of dozens of options while searching for properties. We will implement a feature in which users can type in the neighborhood or community organization into the corresponding dropdown to filter the results.

## Acceptance Criteria

- [ ] When the user clicks on ‘Select Options’ for neighborhood or community organization, in addition to pulling up the dropdown menu, the ‘Select Options’ button will become a text field which can be typed in
- [ ] As a user begins to type, the list of options in the dropdown will filter to only contain the entered text

## Additional context

[https://www.figma.com/design/PAcSR8FjsLtJrxuSMQmOwJ/C%26G-Design-Edits?node-id=0-1&t=qnkbm6RcswNoz1r1-1](url)

![Image](https://github.com/user-attachments/assets/b4fef3ae-b237-4b63-89f4-025a414966b7)

cfreedman:
I'd like to work on this, if possible.

CodeWritingCow:
@cfreedman assigned!

PR #1114, opened 2025-02-26
Key-in filtering for neighborhood search and other filters

cfreedman:
## Description

Updates the existing `MultiSelect` component to allow for key-in input search which filters for the available options for those with matching text. The updated component uses the new `Autocomplete` base component included in the `nextui` library with some adjustments to the styling and behavior to match the existing visuals. In particular, the `Autocomplete` is used as a fully controlled component with selection handled through `onClick` events on the rendered items themselves rather than using the default selection implementation.

Also I updated the `nextui-org/react` dependencies in the lockfile to reach a more recent version that fixes a breaking bug from accessing a property on a `null` value that was cropping up in `Autocomplete`.

## Related Issue(s)

This PR addresses issue #1094 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

I viewed the site locally within the dev server and verified through user interaction that the filters were behaving as expected through responding to input text and displaying the appropriate properties on the map for the selected filters.


cfreedman:
@CodeWritingCow I was hoping to get some design input to finalize this PR concerning the rendering of the selected options chips. Now that the input box is actually being used for text, it didn't seem to appropriate to keep them there and crowd out the user-inputted text.

I currently have them rendering underneath each input box and their container has some height preallocated to accommodate two rows of chips, but this leaves the spacing between the filters looking quite bare when no chips are rendered. But the alternative of dynamic sizing for the chips container leads to content shifting, which also seems not ideal. Let me know if you had any thoughts for about this or a better option, thanks!

nlebovits:
Hey @cfreedman ! Thanks for working on this. @CodeWritingCow has had to step away from the project to focus on work opportunities, so we're in a bit of a leadership transition. @SofiaFasullo is the new lead, and we're working on finding a couple more, so please bear with the slowness here. (I'm also no longer officially on the project.) Can I ask you to make an executive decision here and do what you think is best, since we're shorthanded at the moment?

cfreedman:
No worries, I think that the content shifting ultimately isn't a big deal in this case, so I've just reverted a couple of changes to get it back to that version. It should be good to go, thanks!

nlebovits:
Looks great! Merging. Thanks for working on this.

Issue #1146, opened 2025-03-26
Task: Fix issues identified in `pr_checks_backend`


adamzev:
Following up on #1010, to get to all green checks in `pr_checks_backend` we need to do the following:

For the tests to pass, we need to fix a dockerfile issue related to the password for the postgres container. Since merge requests are submitted from forks, they don't have access to `secrets` so we need to hardcode the password for the ephemeral postgres container. Given that the container is just spun up and discarded this is not a security issue.

For the  `pr_checks_backend`'s system `run-services` to pass, we'll need update the Dockerfile's posgis and timescale versions and eventually fix this more permanently in #1120 .

For the ruff to pass, it looks like we just need to remove an unused variable.
```
new_etl/database.py:83:30: F841 Local variable `transaction` is assigned to but never used
   |
81 |     try:
82 |         # Begin a transaction
83 |         with conn.begin() as transaction:
   |                              ^^^^^^^^^^^ F841
84 |             # Synchronize schema with database table
85 |             if if_exists == "append":
   |
   = help: Remove assignment to unused variable `transaction`
```

PR #1147, opened 2025-03-26
Task: #1146 fix `pr_checks_backend`

adamzev:
# Various small fixes to get `pr_checks_backend` to pass

Fixed dockerfile versions
Hardcoded a temp password for postgres. Given that the container is just spun up and discarded this is not a security issue.
Removed a unused variable for the ruff to pass.


adamzev:
The tests are successfully running now but a new test I recently added isn't mocked correctly for this environment.
`from new_etl.classes.featurelayer import FeatureLayer` imported in `test_metadata_utils` isn't correctly mocking the `bucket = google_cloud_bucket()` which is missing credentials in this env.

cfreedman (approved):

Issue #1125, opened 2025-03-17
Feature Request: Allow the ETL pipeline to run locally without Google/Slack credentials


adamzev:
**Issue:** ETL Pipleline fails with errors regarding Google cloud credentials and Slack credentials. While those are needed for sharing the data, it'd be good to be able to run the pipeline locally without it.

**Workaround:** Add if statements around requests for credentials as uncommited/local code changes.

**Fixes:** Adding settings or checks if those environment variables exist and warning rather than throwing an error

adamzev:
For the bucket, in addition to running locally and with file credentials, we also need the code to run with application default credentials (ADC) in environments like Cloud Run, where workload identity is assigned.

PR #1150, opened 2025-03-28
#1125 Allow non-credentialed users to run ETL pipelines 

adamzev:
# Allow non-credentialed users to run ETL pipelines and support Application Default Credentials 

feat: allow non-credentialed users and cloud env credentials
    
- The ability to specify ports and passwords rather than a full url was added to better suit not having to duplicate secret management (POSTGRES_PASSWORD and VACANT_LOTS_DB)
    
- Added additional logic to psql to user localhost in Cloud Run environments
    
- Added a new bucket manager that should work in the following enviroments:
  - Cloud Run
  - with file credential
  - in read-only mode for non-credentialed users
    
- Changed Slack reporters to not report with a message rather than throwing an error if no credentials exist

## Related Issue(s)

This PR addresses issue #1125 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## How Has This Been Tested?

I've tried this locally with credentials (to my own bucket), without credentials and within Cloud Run. I don't have Slack credentials so I haven't tried it with them.


adamzev:
It looks like the PR Checks Backend are failing just because https://github.com/CodeForPhilly/clean-and-green-philly/pull/1147 isn't in yet.

I want to double check things (there's one Slack message I may have missed) and try this out on different computers before switching it from a draft.

cfreedman (approved):

data/src/config/psql.py
```diff
@@ -1,12 +1,34 @@
...
+# except when running in Cloud Run
+if is_docker() and not is_cloud_run:
+    host = "host.docker.internal"
+else:
+    host = "localhost"
+
+
+if os.getenv("VACANT_LOTS_DB"):
+    # Use the provided database URL
+    url = os.getenv("VACANT_LOTS_DB")
```
adamzev, line 20: I probably should switch this back to 
```
 os.environ["VACANT_LOTS_DB"].replace("localhost", "host.docker.internal")
    if is_docker()
    else os.environ["VACANT_LOTS_DB"]
```
in-case we're in docker and VACANT_LOTS_DB is defined.

data/src/test/test_slack_error_reporter.py
```diff
@@ -5,25 +5,26 @@
 
 sys.path.append(os.path.dirname(os.path.abspath(__file__)) + "/..")
 
-from classes.slack_error_reporter import (
+from new_etl.classes.slack_reporters import (
```
adamzev, line 8: Switch back to classes

adamzev:
I rebased this but saw some things I need to change. I also want to expand this fix to the old ETL pipeline.

PR #1165, opened 2025-04-08
Migrate eslint config to support v9 and update Next dependency

cfreedman:
This PR migrates the eslint config to v9 using the new `eslint.config.js` format as opposed to the old deprecated `eslintrc.json` format. I also added some additional configuration for ignoring generated code in the `.next` directory that shouldn't need to satisfy linting rules that apply only to user-generated code.

Additionally, the `next` dependency was upgraded to `14.2.26` to patch the [CVE-2025-29927](https://github.com/vercel/next.js/security/advisories/GHSA-f82v-jwr5-mffw) vulnerability triggering our pre-commit checks.


adamzev:
Did this pull request not trigger linting checks because it had no code changes? Since eslint changed it may be good to see that those pass so we may want to throw in a no-op change just to trigger that check.

cfreedman:
I looked back and turns out that the frontend `pr_check.yml` workflow is only checking on push not PR opening, so none of the past PRs have been getting linting checks :expressionless: ... I'll make a quick change to that later...

adamzev (approved):

Issue #1016, opened 2024-11-23
Feature Request: Add NDVI and NDBI to ETL pipline


nlebovits:
**Is your feature request related to a problem? Please describe.**
We've discussed adding satellite image-derived NDVI and NDBI to our ETL pipeline to facilitate the identification of vacant land and vacant buildings.

**Describe the solution you'd like**
Using the `pystac` API and Sentinel 2 images, calculate NDVI and NDBI from a composite of the the 3 previous Sentinel 2 images (it has a revisit time of 10 days, so this should roughly correspond to 1 month), using a bounding box for the borders of Philadelphia. Then, extract those NDVI and NDBI raster values to a point on the surface (not a centroid) of all the polygons in our full dataset of the 500k+ properties in Philadelphia (the one created in `new_etl` via `main.py`, not the `script.py` pipeline). 

**Additional context**
Here's Python code that should get you started:

```
# Import required libraries
import odc.stac
import pandas as pd
import planetary_computer
import pystac_client
import xarray as xr
import hvplot.xarray
import panel as pn
from shapely.geometry import box
import geopandas as gpd

# Enable Panel for interactive visualizations
pn.extension()

# Step 1: Connect to Planetary Computer Catalog
catalog = pystac_client.Client.open(
    "https://planetarycomputer.microsoft.com/api/stac/v1",
    modifier=planetary_computer.sign_inplace,  # Automatically signs requests
)

# List available collections
all_collections = [i.id for i in catalog.get_collections()]
sentinel_collections = [collection for collection in all_collections if "sentinel" in collection]
print("Available Sentinel Collections:", sentinel_collections)

# Corrected Query for Sentinel-2 with Cloud Cover Filter
bbox = [-75.2803, 39.8670, -74.9557, 40.1379]  # Philadelphia bounding box
datetime = "2024-06-01/2024-08-31"  # Summer 2024
cloudy_less_than = 10  # Percent cloud cover threshold

search = catalog.search(
    collections=["sentinel-2-l2a"],
    bbox=bbox,
    datetime=datetime,
    query={"eo:cloud_cover": {"lt": cloudy_less_than}}
)
items = search.item_collection()

# Print Results
print(f"Returned {len(items)} Items:")
if items:
    for item in items:
        print(item.id)
else:
    print("No items found with the current filter.")

selected_item = items[0]
selected_item

bands_of_interest = ["red", "green", "blue", "nir", "swir16"]

[...]

PR #1171, opened 2025-04-13
Lebovits/issu 1016 ndvi

nlebovits:
# New Feature: Add NDVI and One-Year NDVI Change to ETL Pipeline

## Description

This PR addresses issue #1016. It uses Planetary Computer to query Sentinel 2 satellite imagery for the most recent summer and the summer prior to that. It returns a raster with the most recent summer's NDVI at 10m resolution, plus the one-year change in NDVI from the prior summer, each as a separate band in the raster. It saves the raster as a file in the `tmp` directory (currently coming in around 65MB), then uses `exactextract` to get the mean NDVI and mean NDVI change for each parcel in the primary feature layer dataset (fall 500k+ properties in Philadelphia). The code is set up to check whether the data for the most recent summer already exists in the `tmp` folder. If it does, it skips straight to extraction. The full code without cached data runs in 10-15 minutes. With the cached data, it takes about 3 minutes. I've also added unit tests in `test/test_ndvi.py`.

### DIFF
--- a/data/src/new_etl/data_utils/ndvi.py
+++ b/data/src/new_etl/data_utils/ndvi.py
@@ -1,0 +1,354 @@
+import os
+import tempfile
+import datetime
+from typing import Tuple
+
+import rioxarray as rxr
+import xarray as xr
+import geopandas as gpd
+import exactextract
+import pystac_client
+import planetary_computer
+import odc.stac
+from shapely.geometry import Polygon, MultiPolygon, box
+
+from ..classes.featurelayer import FeatureLayer
+from ..metadata.metadata_utils import provide_metadata
+from config.config import USE_CRS
+from ..constants.services import CENSUS_BGS_URL
+
+
+def get_current_summer_year() -> int:
+    """
+    Determine the most recent summer year based on current date.
+
+    Summer is defined as June-August. If current month is before June,
+    the most recent summer was last year.
+
+    Returns:
+        int: The year of the most recent summer
+    """
+    today = datetime.datetime.now()
+    if today.month < 6:  # Before June
+        return today.year - 1
+    else:
+        return today.year
+
+
+def get_bbox_from_census_data() -> Tuple[Polygon, Tuple[float, float, float, float]]:
+    """
+    Get bounding box from census block groups data.
+
+    Returns:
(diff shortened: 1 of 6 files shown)


data/src/test/test_ndvi.py
```diff
@@ -0,0 +1,237 @@
...
+            generate_ndvi_data(
+                os.path.join(self.temp_dir, "dummy.tif"),
+                self.test_polygon,
+                (-75.2, 39.9, -75.1, 40.0),
+            )
+        self.assertIn("Test exception", str(context.exception))
+
+    @patch("tempfile.gettempdir")
+    @patch("new_etl.data_utils.ndvi.get_current_summer_year")
+    @patch("new_etl.data_utils.ndvi.get_bbox_from_census_data")
```
adamzev, line 189: it is hard to reason about tests when so much is patched. See if we can eliminate at least 2 of the patches from my suggestions.

data/src/new_etl/data_utils/ndvi.py
```diff
@@ -0,0 +1,355 @@
...
+
+    # Get the GeoDataFrame from the feature layer
+    parcel_gdf = primary_featurelayer.gdf
+
+    # Extract NDVI values to parcels
+    results = extract_ndvi_to_parcels(ndvi_cache_path, parcel_gdf)
+
+    # Update the primary feature layer with the new columns from results
+    # (instead of creating a new one)
+    for column in results.columns:
```
adamzev, line 351: Given that results is on a cleaned version of the gdf and we're adding columns into a version that hasn't been cleaned, is there the possibility for misalignment or different row lengths here when merging in the columns?

It is probably worth making this logic a separate function and trying out some filtering on a df and making sure things merge okay.

adamzev (commented):
I like the number and small size of the helper classes. With our services, I'm planning on moving to a pattern where the service takes the feature layer but then calls to a transform and merge function that just rely on gdfs so they can be tested without worrying about the feature layer class.

I added some suggestions to reduce how much we need to mock and also had a concern regarding whether the row length of `results` could ever be different than `primary_featurelayer.gdf`. 

This should be some very interesting and helpful functionality!

PR #1173, opened 2025-04-13
Feat: add recent activity as ML feature

nlebovits:
# Add recent activity for ML features

## Description
This PR adds permits, business licenses, and appeals data for each property in the dataset, tracking whether there has ever been such a record associated with a given `opa_id`, as well as the date of the most recent record and the days since that date (to the date of the pipeline run). My hunch is that some combination of these features will significantly improve our ability to identify vacant properties with ML.

### DIFF
--- a/data/src/new_etl/data_utils/recent_activity.py
+++ b/data/src/new_etl/data_utils/recent_activity.py
@@ -13,7 +13,6 @@ def recent_activity(primary_featurelayer: FeatureLayer) -> FeatureLayer:
     """
     # Access the GeoDataFrame from the FeatureLayer
     result_gdf = primary_featurelayer.gdf.copy()
-    result_gdf["opa_id"] = pd.to_numeric(result_gdf["opa_id"], errors="coerce")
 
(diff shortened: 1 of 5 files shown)


data/src/new_etl/data_utils/recent_activity.py
```diff
@@ -0,0 +1,62 @@
...
+import requests
+from datetime import datetime, timezone
+
+from ..classes.featurelayer import FeatureLayer
+from ..metadata.metadata_utils import provide_metadata
+from ..constants.services import ACTIVITY_QUERIES
+
+
+def fetch_recent_activity(query: str) -> pd.DataFrame:
+    response = requests.get("https://phl.carto.com/api/v2/sql", params={"q": query})
```
adamzev, line 11: We already should have a class that is capable of grabbing a carto query and converting the results to a df. Is there are reason it didn't work for this case?

adamzev:
Generally, this looks ready to go. I just had one thing I think we should change regarding the carto API call.

PR #1212, opened 2025-05-21
Merging in feature `pipeline-integration` branch

cfreedman:
Reconcile the `pipeline-integration` branch with `staging` both of which contains the main changes with the switch over the updated pipeline, additional data validation checks, etc.

`pipeline-integration` contains the changes regarding removing postgres, switching over to caching for geoparquet files, and some other small changes. We should be able to work off of `staging` moving forward.

There were a number of previous tests relying or earlier postgres patching/old pipeline code still connected to postgres that have been ripped out or left with stubs pending updates with the new unit testing suite to support the new pipeline.


data/src/new_etl/classes/file_manager.py
```diff
@@ -0,0 +1,262 @@
...
+
+        if not os.path.exists(self.temp_directory):
+            os.makedirs(self.temp_directory)
+        if not os.path.exists(self.source_cache_directory):
+            os.makedirs(self.source_cache_directory)
+        if not os.path.exists(self.pipeline_cache_directory):
+            os.makedirs(self.pipeline_cache_directory)
+
+    @staticmethod
+    def get_instance(fraction=CACHE_FRACTION):
```
grahamalama, line 66: I think this could also be a part of the `__init__` method, right? Like 

```
if FileManager._instance is not None:
    return _instance
```

that way you'd avoid the need for `get_instance` and raising a `ValueError`, and callers don't need to know that `FileManager` is a singleton.

data/src/new_etl/classes/file_manager.py
grahamalama, line 1: This seems like an important class for the new ETL pipeline. Does it deserve any accompanying tests in this PR?

data/src/main.py
```diff
@@ -167,37 +164,64 @@
...
+    # to_postgis_with_schema(dataset.gdf, "all_properties_end", conn)
 
     # Generate and send diff report
-    diff_report = DiffReport()
-    diff_report.run()
+    # diff_report = DiffReport()
+    # diff_report.run()
 
-    send_pg_stats_to_slack(conn)  # Send PostgreSQL stats to Slack
+    # send_pg_stats_to_slack(conn)  # Send PostgreSQL stats to Slack
```
grahamalama, line 211: Do we need this commented out code, or can we safely delete it?

grahamalama (commented):
Nice work! Removing big dependencies like Postgres will really make reasoning about the project easier.

Because we ran the command together to completion IRL at a hack night, I didn't look _too_ closely at the internals of each new or modified function, since there's quite a bit changing in this PR. 

I did notice some commented-out code, explicitly passing tests, and tests missing for new code. I think it'd be good to square that stuff away before merging this PR.

data/src/test/test_data_utils.py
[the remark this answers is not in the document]
cfreedman: These were largely dependent on the previous postgres implementation or some heavy mocking of it - I'm keeping them stubbed out for now until we fill them in with the planned unit testing PR coming after the `FeatureLayer` refactor, which will change how all the services are called.

data/src/main.py
cfreedman: Yep we can remove that for now.

data/src/new_etl/data_utils/vacant_properties.py
```diff
@@ -19,10 +19,12 @@ def load_backup_data_from_gcs(file_name: str) -> pd.DataFrame:
         pd.DataFrame: A DataFrame containing the backup data with only the "opa_id" column.
     """
     bucket = google_cloud_bucket()
+    if not bucket:
+        print("No Google Cloud bucket available - skipping backup data load.")
+        return None
```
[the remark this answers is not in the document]
cfreedman: Makes sense to me - I think we should still allow for people to run the pipeline locally without Google cloud credentials setup to load the vacancies data (or possibly in the future we could some sort of fallback solution), so we'd still need to catch the exception somewhere. Maybe just one level up where this is being called in `vacant_properties`?

data/src/new_etl/data_utils/vacant_properties.py
```diff
@@ -97,25 +99,26 @@ def vacant_properties(primary_featurelayer: FeatureLayer) -> FeatureLayer:
...
+
+    #     if backup_gdf:
+    #         # Add parcel_type column to backup data
+    #         backup_gdf["parcel_type"] = "Land"
+
+    #         # Append backup data to the existing dataset
+    #         print(f"Appending backup data ({len(backup_gdf)} rows) to the existing data.")
+    #         vacant_properties.gdf = pd.concat(
+    #             [vacant_properties.gdf, backup_gdf], ignore_index=True
+    #         )
```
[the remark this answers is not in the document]
cfreedman: Yea, need that put back in for correct loading of the vacancy data backups, thanks

data/src/new_etl/data_utils/vacant_properties.py
```diff
@@ -105,17 +107,26 @@ def vacant_properties(primary_featurelayer: FeatureLayer) -> FeatureLayer:
...
-            [vacant_properties.gdf, backup_gdf], ignore_index=True
-        )
+            # Append backup data to the existing dataset
+            print(
+                f"Appending backup data ({len(backup_gdf)} rows) to the existing data."
+            )
+            vacant_properties.gdf = pd.concat(
+                [vacant_properties.gdf, backup_gdf], ignore_index=True
+            )
+        except Exception as e:
```
grahamalama, line 126: Something to add an issue for -- we probably want to add an issue to refactor these `except Exception as e` lines. I think we generally want to try to be more specific with error catching, as it usually helps debug issues. Not something to address in this PR though.

data/src/test/test_data_utils.py
grahamalama, line 1: Instead of using `pass`, can we use `@pytest.mark.skip` for these tests? If we intend to come back to them, this will at least create a bit more noise in CI output than simply `pass`ing them.

data/src/new_etl/classes/file_manager.py
grahamalama: Talked with @cfreedman on Slack, and we agreed to postpone tests for this class to unblock other work around this area. Could we create an issue so that we remember to come back and add tests for this?

PR #1213, opened 2025-05-22
Refactor slack reporting service

cfreedman:
Changes the slack reporting to be managed by a class that initializes a single client connection. The main methods involve providing summarizing statistics from the `all_properties_end` final dataframe as well as metadata information about all the generated parquet files from the pipeline and sending it to Slack.

There's also some preliminary changes to the datadiffing that reports the changes the most recent two parquet files (if they exist), but this will need to be further refactored in the future to allow for a more flexible approach for calling data-diffing at each service stage in the pipeline upon the raw, sourced data.


data/src/new_etl/classes/slack_reporters.py
grahamalama, line 1: Should we add some tests for this new class?

grahamalama (commented):
Looks good overall! One minor question, but I do think it'd be worth adding some tests for this class while we're here.

Also, random thought, but if we wanted to get fancy, I bet we could set this up as a custom python logger somehow. Here's an [example](https://github.com/chinnichaitanya/python-slack-logger/blob/master/slack_logger/message_logger.py) from a different project. _Way_ out of scope for this PR, but maybe something worth adding to the backlog. 

One benefit that'd have would be that you wouldn't need to check if the logger was initialized before using it. I think we could configure it so that we call the logger, and it only attempts to actually send the message to Slack if everything is configured correctly.

data/src/main.py
```diff
@@ -198,17 +224,21 @@
         str
     )
 
-    # Dataset profiling
-    # send_dataframe_profile_to_slack(dataset.gdf, "all_properties_end")
-
-    # Save dataset to PostgreSQL
-    # to_postgis_with_schema(dataset.gdf, "all_properties_end", conn)
+    token = os.getenv("CAGP_SLACK_API_TOKEN")
```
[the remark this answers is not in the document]
cfreedman: Just missed some of the duplicated code, thanks!

data/src/new_etl/classes/slack_reporters.py
grahamalama: Sounds goo.d Can we create an issue so we don't lose track of that work?

grahamalama (approved):

PR #1256, opened 2025-06-29
Revert earlier commit with validators

cfreedman:
Reverting earlier commit to add in more of the output validators at a later point.


nlebovits:
No need to revert now. I rebased and resolved it.

